PR #352, opened 2022-12-22
Add license: GPLv3 for the program, MIT for everything else

ckamm:


ckamm:
will be done differently

PR #353, opened 2022-12-27
reduce only order type and mode for tokens and perp markets

microwavedcola1:


programs/mango-v4/src/instructions/perp_place_order.rs
```diff
@@ -109,8 +110,32 @@ pub fn perp_place_order(ctx: Context<PerpPlaceOrder>, order: Order, limit: u8) -
 
     let now_ts: u64 = Clock::get()?.unix_timestamp.try_into().unwrap();
 
-    // TODO apply reduce_only flag to compute final base_lots, also process event queue
-    require!(order.reduce_only == false, MangoError::SomeError);
+    let effective_pos = account
+        .perp_position(perp_market_index)?
+        .effective_base_position_lots();
+    if perp_market.is_reduce_only() {
+        require!(
```
microwavedcola1, line 117: hmm I saw something similar in v3, basically for spot orders, we restricted user to be able to have atmost 1 open order after market went into reduce only, wdyt?

programs/mango-v4/src/instructions/serum3_place_order.rs
```diff
@@ -204,7 +204,7 @@ pub fn serum3_place_order(
     ctx: Context<Serum3PlaceOrder>,
     side: Serum3Side,
     limit_price: u64,
-    max_base_qty: u64,
+    mut max_base_qty: u64,
```
microwavedcola1, line 207: I feel weird about mut'ing this here, wdyt?

programs/mango-v4/src/instructions/perp_place_order.rs
```diff
@@ -38,7 +39,7 @@ pub struct PerpPlaceOrder<'info> {
 
 // TODO
 #[allow(clippy::too_many_arguments)]
-pub fn perp_place_order(ctx: Context<PerpPlaceOrder>, order: Order, limit: u8) -> Result<()> {
+pub fn perp_place_order(ctx: Context<PerpPlaceOrder>, mut order: Order, limit: u8) -> Result<()> {
```
microwavedcola1, line 42: I feel weird about mut'ing this here, wdyt?

programs/mango-v4/src/instructions/serum3_place_order.rs
```diff
@@ -204,7 +204,7 @@ pub fn serum3_place_order(
     ctx: Context<Serum3PlaceOrder>,
     side: Serum3Side,
```
microwavedcola1, line 205: I have not added a reduce_only order type for serum, only market can have reduce only, wdyt?

programs/mango-v4/src/state/mango_account_components.rs
```diff
@@ -284,6 +284,10 @@ impl PerpPosition {
         self.base_position_lots
     }
 
+    pub fn effective_base_position_lots(&self) -> i64 {
+        self.base_position_lots + self.taker_base_lots
```
microwavedcola1, line 288: this ignores what is in orders, because I enforce only one order at a time, wdyt?

programs/mango-v4/src/instructions/perp_place_order.rs
```diff
@@ -109,8 +110,22 @@ pub fn perp_place_order(ctx: Context<PerpPlaceOrder>, order: Order, limit: u8) -
...
-    // TODO apply reduce_only flag to compute final base_lots, also process event queue
-    require!(order.reduce_only == false, MangoError::SomeError);
+    let effective_pos = account
+        .perp_position(perp_market_index)?
+        .effective_base_position_lots();
+    order.max_base_lots = if order.reduce_only || perp_market.is_reduce_only() {
+        if (order.side == Side::Bid && effective_pos > 0)
+            || (order.side == Side::Ask && effective_pos < 0)
+        {
+            0
```
ckamm, line 120: This is another place where we can debate error vs do-nothing-success. 

- If the market is reduce-only and the user sends a non-reduce-only order, wouldn't we expect an error here?
- If the user sends a reduce-only order, then silent success is what I'd expect.

A similar thing applies to the reduction of the order max_base_lots below, imo. If the user didn't request reduce-only, then I'd rather error instead of silently reducing their order amount.

programs/mango-v4/src/instructions/perp_place_order.rs
```diff
@@ -109,8 +110,22 @@ pub fn perp_place_order(ctx: Context<PerpPlaceOrder>, order: Order, limit: u8) -
 
     let now_ts: u64 = Clock::get()?.unix_timestamp.try_into().unwrap();
 
-    // TODO apply reduce_only flag to compute final base_lots, also process event queue
-    require!(order.reduce_only == false, MangoError::SomeError);
+    let effective_pos = account
+        .perp_position(perp_market_index)?
+        .effective_base_position_lots();
```
ckamm, line 115: imo this should take into account orders the user already has on the book. If the user has effective_pos 5 plus ask_lots = 4, they should only be able to add an ask of size 1.

It's unclear how bid_lots would affect this. Ignore them?

programs/mango-v4/src/instructions/serum3_edit_market.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub admin: Signer<'info>,
+
+    #[account(
+        mut,
+        has_one = group
+    )]
+    pub market: AccountLoader<'info, Serum3Market>,
+}
+
+pub fn serum3_edit_market(ctx: Context<Serum3EditMarket>, reduce_only: bool) -> Result<()> {
```
ckamm, line 22: since this is not unlikely to expand in the future, use the usual *_edit pattern where we pass Options

programs/mango-v4/src/instructions/serum3_place_order.rs
ckamm: ok with it

programs/mango-v4/src/instructions/serum3_place_order.rs
```diff
@@ -258,6 +258,48 @@ pub fn serum3_place_order(
...
+            .native(&bank)
+            .to_num::<i64>();
+        let base_lot_size = load_market_state(
+            &ctx.accounts.serum_market_external,
+            &ctx.accounts.serum_program.key(),
+        )?
+        .coin_lot_size;
+        let max_base_qty_native = max_base_qty * base_lot_size;
+        require!(
+            max_base_qty_native >= token_pos_native.unsigned_abs(),
```
ckamm, line 279: I don't get this condition. Sounds like if the user has 10 base native and they want to buy 9 more that's an error, but buying 11 is fine? Can you explain?

programs/mango-v4/src/instructions/serum3_place_order.rs
```diff
@@ -258,6 +258,48 @@ pub fn serum3_place_order(
...
+        require!(
+            load_open_orders_ref(&ctx.accounts.open_orders.as_ref())?.free_slot_bits == u128::MAX,
+            MangoError::SomeError
+        );
+
+        let bank = ctx.accounts.payer_bank.load()?;
+        let token_pos_native = account
+            .token_position(bank.token_index)?
+            .native(&bank)
+            .to_num::<i64>();
```
ckamm, line 271: potential rounding issue, -9.1 would be -10

programs/mango-v4/src/instructions/serum3_edit_market.rs
```diff
@@ -0,0 +1,25 @@
...
+    ctx.accounts.market.load_mut()?.reduce_only = u8::from(reduce_only);
```
ckamm, line 23: What does it mean for a serum market to be reduce-only?

It makes sense to me for perp-markets to be reduce-only: users must reduce their perp base position. Banks can also be reduce-only: users can't increase borrows or increase deposits.

But spot markets have no state that could be reduced? I'd say we need a "can't place new orders on this market" flag, but that should be all?

programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -100,18 +100,30 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
...
         remaining_accounts: &[AccountInfo],
         amount: u64,
+        reduce_only: bool,
         allow_token_account_closure: bool,
     ) -> Result<()> {
         require_msg!(amount > 0, "deposit amount must be positive");
 
         let token_index = self.bank.load()?.token_index;
 
+        let bank = self.bank.load()?;
```
ckamm, line 110: load the bank first, then you don't have to re-load it for the token_index

programs/mango-v4/src/instructions/serum3_place_order.rs
```diff
@@ -258,6 +258,48 @@ pub fn serum3_place_order(
...
+                if token_pos_native < 0 {
+                    0
+                } else {
+                    (token_pos_native.max(0) as u64).min(max_base_qty)
+                }
+            }
+        }
+    } else {
+        max_base_qty
+    };
```
ckamm, line 301: Instead of this logic, I'd make it dependent on bank.is_reduce_only. Every place where we check the deposit_to_vault_ratio likely also needs to care about the bank's reduce-only state.

programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -100,18 +100,30 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
...
 
         let (position, raw_token_index) = account.token_position_mut(token_index)?;
 
-        let amount_i80f48 = I80F48::from(amount);
+        let amount_i80f48 = if reduce_only || bank.is_reduce_only() {
+            position
+                .native(&bank)
+                .min(I80F48::ZERO)
+                .abs()
+                .min(I80F48::from(amount))
```
ckamm, line 122: same "error vs silent reduction" thing as for perps

programs/mango-v4/src/instructions/token_register_trustless.rs
```diff
@@ -99,10 +105,10 @@ pub fn token_register_trustless(
...
         loan_origination_fee_rate: I80F48::from_num(0.0005),
         loan_fee_rate: I80F48::from_num(0.005),
-        maint_asset_weight: I80F48::from_num(0),
-        init_asset_weight: I80F48::from_num(0),
-        maint_liab_weight: I80F48::from_num(1.4), // 2.5x
-        init_liab_weight: I80F48::from_num(1.8),  // 1.25x
+        maint_asset_weight: I80F48::from_num(0), // 2.5x
+        init_asset_weight: I80F48::from_num(0),  // 1.25x
+        maint_liab_weight: I80F48::from_num(1.4),
+        init_liab_weight: I80F48::from_num(1.8),
```
ckamm, line 111: moving these comments is probably an accident?

programs/mango-v4/src/instructions/token_withdraw.rs
```diff
@@ -59,8 +59,13 @@ pub fn token_withdraw(ctx: Context<TokenWithdraw>, amount: u64, allow_borrow: bo
     require_msg!(amount > 0, "withdraw amount must be positive");
 
     let group = ctx.accounts.group.load()?;
+    let mut bank = ctx.accounts.bank.load_mut()?;
     let token_index = ctx.accounts.bank.load()?.token_index;
```
ckamm, line 63: drop that extra bank.load()

programs/mango-v4/src/instructions/token_withdraw.rs
```diff
@@ -59,8 +59,13 @@ pub fn token_withdraw(ctx: Context<TokenWithdraw>, amount: u64, allow_borrow: bo
...
 
+    if bank.is_reduce_only() {
+        require!(!allow_borrow, MangoError::SomeError);
+    }
```
ckamm, line 67: we should allow withdraws that don't trigger borrows, even if allow_borrow is true

programs/mango-v4/src/state/mango_account.rs
```diff
@@ -505,6 +505,12 @@ impl<
         (0..self.header().perp_oo_count()).map(|i| self.perp_order_by_raw_index(i))
     }
 
+    pub fn any_perp_orders(&self, perp_market: &PerpMarket) -> bool {
```
ckamm, line 508: unused?

programs/mango-v4/src/state/mango_account_components.rs
ckamm: I don't see "only one order" being enforced. Can you add a comment saying that this takes into account base lots from unprocessed events, but not anything from open orders?

microwavedcola1: I thought we agreed on not enforcing - only one order - when placing a new perp order on a perp market in reduce only mode, or?

programs/mango-v4/src/instructions/perp_place_order.rs
microwavedcola1: Done! Please re-review flow in place perp order ix

programs/mango-v4/src/instructions/serum3_edit_market.rs
microwavedcola1: done, please check

programs/mango-v4/src/instructions/serum3_place_order.rs
microwavedcola1: sorry it should be, 

```
max_base_qty_native <= token_pos_native.unsigned_abs()
```

I have fixed it now, intention is, that you cant place an order greater than your position, when market is in reduce only mode, see https://github.com/blockworks-foundation/mango-v3/blob/main/program/src/processor.rs#L2177

programs/mango-v4/src/instructions/serum3_edit_market.rs
microwavedcola1: You have a good point, but how does one close deposits/borrows on token positions? If I remove my current serum market reduce only mode, alternative to closing token positions would only be to deposit/withdraw

programs/mango-v4/src/instructions/token_deposit.rs
microwavedcola1: done

programs/mango-v4/src/instructions/token_register_trustless.rs
microwavedcola1: good catch, fixed, I did the vscode struct field reordering, apparently it didnt take the comments with it

programs/mango-v4/src/instructions/token_withdraw.rs
microwavedcola1: done

programs/mango-v4/src/instructions/serum3_place_order.rs
microwavedcola1: Good point, I can actually drop the serum3_market.is_reduce_only completely, and only rely on corresponding banks reduce only. 

Another thought, I have not even thought about what it means if quote token is non usd for serum market hmm.

I was also thinking about deposit_to_vault_ratio, and I think it should not apply when bank has reduce only mode.

programs/mango-v4/src/instructions/serum3_place_order.rs
microwavedcola1: hmm would you recommend always flooring?

programs/mango-v4/src/instructions/perp_place_order.rs
microwavedcola1: hmm agreed, I have done something like this now
image

programs/mango-v4/src/state/mango_account_components.rs
microwavedcola1: ok as discussed there would be no enforcing

programs/mango-v4/src/instructions/perp_place_order.rs
microwavedcola1: please check latest change

programs/mango-v4/src/instructions/serum3_place_order.rs
microwavedcola1: as discussed this whole block is removed

programs/mango-v4/src/instructions/serum3_edit_market.rs
microwavedcola1: as discussed serum reduce only is very simple, doesnt support placing new orders

programs/mango-v4/src/instructions/serum3_place_order.rs
microwavedcola1: as discussed,  deposit_to_vault_ratio is independent of reduce only mode

programs/mango-v4/src/instructions/flash_loan.rs
```diff
@@ -377,6 +377,10 @@ pub fn flash_loan_end<'key, 'accounts, 'remaining, 'info>(
     let mut token_loan_details = Vec::with_capacity(changes.len());
     for (change, oracle_price) in changes.iter().zip(oracle_prices.iter()) {
         let mut bank = health_ais[change.bank_index].load_mut::<Bank>()?;
+        if change.amount.abs() > 0 {
+            require!(!bank.is_reduce_only(), MangoError::SomeError);
+        }
```
ckamm, line 382: I don't think this check is right. It says: any change at all is forbidden for reduce_only banks. But using flash loan to repay borrows or withdraw deposits should be allowed!

programs/mango-v4/src/instructions/perp_place_order.rs
```diff
@@ -109,8 +110,31 @@ pub fn perp_place_order(ctx: Context<PerpPlaceOrder>, order: Order, limit: u8) -
 
     let now_ts: u64 = Clock::get()?.unix_timestamp.try_into().unwrap();
 
-    // TODO apply reduce_only flag to compute final base_lots, also process event queue
-    require!(order.reduce_only == false, MangoError::SomeError);
+    let pp = account.perp_position(perp_market_index)?;
+    let mut effective_pos = pp.effective_base_position_lots();
+    effective_pos = if effective_pos >= 0 {
+        effective_pos - pp.asks_base_lots
```
ckamm, line 116: Add a comment with the reasoning for including open orders in one direction and event queue lots. 

Shouldn't the casing depend on `side`? Consider trying to place a bid when you have 10 base_lots and 15 lots in open asks. Right now the effective_pos would be -5, meaning we could add 5 more bids? That would be incorrect.

Also, I'd prefer a .max(0) here and a .min(0) below. 

Finally, maybe this should be called the base_lots_limit and only be computed in the `if reduce_only` case?

programs/mango-v4/src/instructions/perp_place_order.rs
ckamm: Almost. I'm thinking of integrators that want to close their position and forget to send reduce-only. In reduce-only markets, their order would now be an error, even if it would work without adjustment.

I think for best behavior, we should accept orders that are not tagged reduce-only even when the market is reduce-only -- but only as long as they can go though unadjusted.

programs/mango-v4/src/instructions/serum3_place_order.rs
```diff
@@ -212,6 +212,7 @@ pub fn serum3_place_order(
     limit: u16,
 ) -> Result<()> {
     let serum_market = ctx.accounts.serum_market.load()?;
+    require!(!serum_market.is_reduce_only(), MangoError::SomeError);
```
ckamm, line 215: Deserves a real error message, since users/integrators are likely to see it.

programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -100,20 +100,38 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
...
         require_msg!(amount > 0, "deposit amount must be positive");
 
-        let token_index = self.bank.load()?.token_index;
+        let mut bank = self.bank.load_mut()?;
+        let token_index = bank.token_index;
+
+        let (amount_i80f48, token_index) = {
+            // Get the account's position for that token index
+            let account = self.account.load()?;
+            let position = account.token_position(token_index)?;
```
ckamm, line 114: immediately after this block we load account/position again. Could just move the mut-loading above this block to save cu?

programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -100,20 +100,38 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
...
+
+            if bank.is_reduce_only() {
+                require!(reduce_only, MangoError::SomeError);
```
ckamm, line 117: similar to perp: I think we should allow non-reduce-only requests to go through as long as they are compatible with the reduce-only bank

programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -100,20 +100,38 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
...
+        reduce_only: bool,
         allow_token_account_closure: bool,
     ) -> Result<()> {
         require_msg!(amount > 0, "deposit amount must be positive");
 
-        let token_index = self.bank.load()?.token_index;
+        let mut bank = self.bank.load_mut()?;
+        let token_index = bank.token_index;
+
+        let (amount_i80f48, token_index) = {
```
ckamm, line 111: why's token_index returned again?

programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -100,20 +100,38 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
...
+                position
+                    .native(&bank)
+                    .min(I80F48::ZERO)
+                    .abs()
+                    .min(I80F48::from(amount))
+            } else {
+                I80F48::from(amount)
+            };
+            (amount_i80f48, token_index)
+        };
```
ckamm, line 129: need to take care here, amount_i80f48 is potentially 0

programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -122,10 +140,12 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
         };
 
         // Transfer the actual tokens
-        token::transfer(self.transfer_ctx(), amount)?;
+        token::transfer(
+            self.transfer_ctx(),
+            amount_i80f48.round_to_zero().to_num::<u64>(),
```
ckamm, line 145: This rounding costs dust and is potentially dangerous. I'd prefer if we allowed -4.7 native tokens to be repaid with 5 native tokens (already compute that in the block where you compute `amount`). For tokens that are not marked in-use, the dusting mechanism would then coerce the user's remaining 0.3 native tokens to 0.

programs/mango-v4/src/state/serum3_market.rs
```diff
@@ -28,11 +28,13 @@ pub struct Serum3Market {
 
     pub registration_time: u64,
 
-    pub reserved: [u8; 128],
+    pub reduce_only: u8,
```
ckamm, line 31: put it into padding1 instead, imo

ckamm (commented):
The new functionality is missing tests.

programs/mango-v4/src/instructions/perp_place_order.rs
microwavedcola1: Done

programs/mango-v4/src/instructions/flash_loan.rs
microwavedcola1: done

programs/mango-v4/src/instructions/token_deposit.rs
microwavedcola1: done

programs/mango-v4/src/instructions/token_deposit.rs
microwavedcola1: q: why would 0 need special handling?

programs/mango-v4/src/instructions/flash_loan.rs
ckamm: Suggestion for reduce-only:
- if change < 0 && native + change < 0: error
- if change > 0 && native + change >= 1: error

programs/mango-v4/src/instructions/flash_loan.rs
```diff
@@ -391,18 +389,31 @@ pub fn flash_loan_end<'key, 'accounts, 'remaining, 'info>(
...
         };
 
+        let loan_origination_fee = cm!(loan * bank.loan_origination_fee_rate);
+        cm!(bank.collected_fees_native += loan_origination_fee);
+
+        if bank.is_reduce_only() {
+            let change_amount = if change.amount >= 0 {
+                cm!(change.amount - loan_origination_fee)
+            } else {
+                cm!(change.amount + loan_origination_fee)
```
ckamm, line 399: this looks wrong, shouldn't the origination fee always be subtracted, regardless of the sign of `change.amount`?

programs/mango-v4/src/instructions/flash_loan.rs
```diff
@@ -391,18 +389,31 @@ pub fn flash_loan_end<'key, 'accounts, 'remaining, 'info>(
...
+            require!(
+                (change_amount < 0 && native + change_amount >= 0)
+                    || (change_amount > 0 && native + change_amount < 2),
+                MangoError::SomeError
+            );
+        }
+
         // Enforce min vault to deposits ratio
-        if loan > 0 {
+        if native + change.amount - loan_origination_fee < 0 {
```
ckamm, line 409: cm, and since we also need it inside the `if reduce-only`, maybe pre-compute it before that block

programs/mango-v4/src/instructions/flash_loan.rs
```diff
@@ -391,18 +389,31 @@ pub fn flash_loan_end<'key, 'accounts, 'remaining, 'info>(
...
+
+        if bank.is_reduce_only() {
+            let change_amount = if change.amount >= 0 {
+                cm!(change.amount - loan_origination_fee)
+            } else {
+                cm!(change.amount + loan_origination_fee)
+            };
+            require!(
+                (change_amount < 0 && native + change_amount >= 0)
+                    || (change_amount > 0 && native + change_amount < 2),
```
ckamm, line 403: should be <1 I'd say

programs/mango-v4/src/instructions/perp_place_order.rs
```diff
@@ -111,30 +111,34 @@ pub fn perp_place_order(ctx: Context<PerpPlaceOrder>, mut order: Order, limit: u
...
-    if perp_market.is_reduce_only() {
-        require!(order.reduce_only, MangoError::SomeError)
-    };
-
-    order.max_base_lots = if order.reduce_only {
-        if (order.side == Side::Bid && effective_pos > 0)
+    let effective_pos = pp.effective_base_position_lots();
+    let max_base_lots = if order.reduce_only || perp_market.is_reduce_only() {
+        if (order.side == Side::Bid && effective_pos >= 0)
             || (order.side == Side::Ask && effective_pos < 0)
```
ckamm, line 117: if the bids line gets >=, this should get <=

programs/mango-v4/src/instructions/token_deposit.rs
ckamm: it was just a reminder to myself; I think spl transfers of 0 amount will succeed

programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -140,10 +144,7 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
         };
 
         // Transfer the actual tokens
-        token::transfer(
-            self.transfer_ctx(),
-            amount_i80f48.round_to_zero().to_num::<u64>(),
-        )?;
+        token::transfer(self.transfer_ctx(), amount_i80f48.to_num::<u64>())?;
```
ckamm, line 147: this to_num() is a bit unfortunate, but should always succeed. I think this might possibly be rewritten such that we have an integer amount here instead of the I80F48 (convert to integer near the .abs().ceil(), but even that is a bit awkward; I recently made a `I80F48::clamp_to_u64()` for situations like this in another branch)

programs/mango-v4/src/instructions/flash_loan.rs
```diff
@@ -392,21 +392,18 @@ pub fn flash_loan_end<'key, 'accounts, 'remaining, 'info>(
...
-                cm!(change.amount - loan_origination_fee)
-            } else {
-                cm!(change.amount + loan_origination_fee)
-            };
             require!(
-                (change_amount < 0 && native + change_amount >= 0)
-                    || (change_amount > 0 && native + change_amount < 2),
+                (change_amount < 0 && native_after_change >= 0)
+                    || (change_amount > 0 && native_after_change < 1),
                 MangoError::SomeError
```
ckamm, line 401: minor: Bank is reduce-only error

programs/mango-v4/tests/program_test/mango_client.rs
```diff
@@ -617,7 +617,7 @@ impl ClientInstruction for TokenWithdrawInstruction {
 
 pub struct TokenDepositInstruction {
```
ckamm, line 618: possibly we should make a `SimpleTokenDepositInstruction` that doesn't need the reduce_only and bank_index args. Or maybe we implement defaults and do `...Default::default()` everywhere. But not in this PR I'd say.

programs/mango-v4/tests/program_test/mango_client.rs
```diff
@@ -1160,6 +1163,82 @@ impl ClientInstruction for TokenResetNetBorrows {
...
+            stable_price_delay_growth_limit_opt: None,
+            stable_price_growth_limit_opt: None,
+            min_vault_to_deposits_ratio_opt: None,
+            net_borrow_limit_per_window_quote_opt: None,
+            net_borrow_limit_window_size_ts_opt: None,
+            borrow_weight_scale_start_quote_opt: None,
+            deposit_weight_scale_start_quote_opt: None,
+            reset_stable_price: false,
+            reset_net_borrow_limit: false,
+            reduce_only_opt: Some(true),
```
ckamm, line 1241: does the instruction implement Default? probably not. If we need to do this in more than one place it may be good to add it.

ts/client/src/mango_v4.ts
```diff
@@ -4506,12 +4562,16 @@ export type MangoV4 = {
...
+            "name": "reduceOnly",
+            "type": "u8"
+          },
           {
             "name": "reserved",
             "type": {
               "array": [
                 "u8",
-                128
+                127
```
ckamm, line 4574: idl needs regeneration

programs/mango-v4/tests/test_reduce_only.rs
```diff
@@ -0,0 +1,571 @@
...
+            owner,
+            token_account: payer_mint_accounts[0],
+            token_authority: payer,
+            bank_index: 0,
+        },
+    )
+    .await;
+    assert!(res.is_err());
+
+    // deposit with reduce_only should pass silently
```
ckamm, line 81: nit: ... with no effect

programs/mango-v4/tests/test_reduce_only.rs
```diff
@@ -0,0 +1,571 @@
...
+        },
+    )
+    .await
+    .unwrap();
+
+    // borrowing should fail
+    let res = send_tx(
+        solana,
+        TokenWithdrawInstruction {
+            amount: initial_token_deposit,
```
ckamm, line 120: maybe just borrow 1 - possibly there isn't enough collateral to borrow this much

programs/mango-v4/tests/test_reduce_only.rs
```diff
@@ -0,0 +1,571 @@
...
+            allow_borrow: true,
+            account: account_0,
+            owner,
+            token_account: payer_mint_accounts[0],
+            bank_index: 0,
+        },
+    )
+    .await;
+    assert!(res.is_err());
```
ckamm, line 129: could you add a simple test for borrow repayment with reduce-only?

programs/mango-v4/tests/test_reduce_only.rs
```diff
@@ -0,0 +1,571 @@
...
+            quote_lot_size: 10,
+            base_lot_size: 100,
+            maint_asset_weight: 0.975,
+            init_asset_weight: 0.95,
+            maint_liab_weight: 1.025,
+            init_liab_weight: 1.05,
+            liquidation_fee: 0.012,
+            maker_fee: 0.0002,
+            taker_fee: 0.000,
+            settle_pnl_limit_factor: 0.2,
```
ckamm, line 205: highly recommend you set this to -1 in tests that don't explicitly test settle limits, otherwise they become pretty complex (this test is fine, doesn't settle anything after all)

programs/mango-v4/tests/test_reduce_only.rs
```diff
@@ -0,0 +1,571 @@
...
+
+    let price_lots = {
+        let perp_market = solana.get_account::<PerpMarket>(perp_market).await;
+        perp_market.native_price_to_lot(I80F48::from(1000))
+    };
+
+    // Set the initial oracle price
+    send_tx(
+        solana,
+        StubOracleSetInstruction {
```
ckamm, line 221: generally prefer using set_perp_stub_oracle_price because it also resets stable price (and is shorter to write, too)

programs/mango-v4/tests/test_reduce_only.rs
```diff
@@ -0,0 +1,571 @@
...
+    let res = send_tx(
+        solana,
+        PerpPlaceOrderInstruction {
+            account: account_0,
+            perp_market,
+            owner,
+            side: Side::Bid,
+            price_lots: {
+                let perp_market = solana.get_account::<PerpMarket>(perp_market).await;
+                perp_market.native_price_to_lot(I80F48::from(500))
```
ckamm, line 290: maybe use `price_lots/2`?

programs/mango-v4/tests/test_reduce_only.rs
```diff
@@ -0,0 +1,571 @@
...
+            max_base_lots: 1,
+            max_quote_lots: i64::MAX,
+            reduce_only: false,
+            client_order_id: 0,
+        },
+    )
+    .await;
+    assert!(res.is_err());
+
+    // account_0 - place a new bid with reduce only true should pass
```
ckamm, line 352: ...and do nothing

PR #356, opened 2022-12-30
rs CI improvements

silas-x:
### Summary
Simplifies the Rust CI job and removes dependency on unmaintained action-rs steps to install toolchain. The changes should speed up the job without impacting tests.

### Detail
- Removed working dir as there is no need to specify it
- Removed step to install Linux deps as the runner doesn't need them anymore
~~- Removed toolchain install step as the runner comes with latest -1 version of rust (currently 1.65). The `action-rs` repo is no longer maintained and uses deprecated steps. _If we want to test on various specific toolchain versions we can build something better - let me know._~~
- Added step to check/log rust version for troubleshooting purposes
- Added cache for the 2nd job
- Removed `cargo test` step as the only tests it was running was for anchor cli


.github/workflows/ci-code-review-rust.yml
```diff
@@ -41,20 +36,12 @@ jobs:
...
-        run: sudo apt-get update && sudo apt-get install -y pkg-config build-essential libudev-dev
-
-      - name: Install Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          override: true
-          profile: minimal
-          toolchain: ${{ env.RUST_TOOLCHAIN }}
-          components: rustfmt, clippy
-
```
ckamm, line 54: @silas-x How does this work? We need an action to install the rust toolchain, right?

ckamm:
clippy seems to fail now that it's upgraded. I had the same issue this morning and added some --allow flags to get clippy past failing on anchor: https://github.com/blockworks-foundation/mango-v4/pull/357/commits/0bad232315b6c5ddf420d728dfcb13c18a0fb50f

silas-x:
Looks like they're struggling to upgrade toolchain version in the Anchor repo ([https://github.com/coral-xyz/anchor/pull/2272](url)).

Added back in toolchain version, but without using the `action-rs` action. Included and tested the allow args you mentioned. Can revisit if they're needed when we upgrade to anchor 0.26.

silas-x:
@ckamm if you're OK with this, I can resolve the conflict and merge. Should shave off 6-8mins off testing time

ckamm:
Let's squash-merge once CI is good

PR #357, opened 2023-01-02
rust client: Make AccountFetcher futures Send

ckamm:
This required redoing the cached account fetcher logic to properly deal with locking and repeated calls while a fetch is ongoing.


microwavedcola1 (approved):

riordanp (approved):
Working well for my use case

PR #362, opened 2023-01-04
combine open orders create with place order

abrzezinski94:


ts/client/src/client.ts
```diff
@@ -1141,18 +1193,49 @@ export class MangoClient {
...
+
+      const [openOrderPublicKey] = await PublicKey.findProgramAddress(
+        [
+          Buffer.from('Serum3OO'),
+          mangoAccount.publicKey.toBuffer(),
+          serum3Market.publicKey.toBuffer(),
+        ],
+        this.program.programId,
+      );
+      const baseBank = group.getFirstBankByTokenIndex(
```
abrzezinski94, line 1218: Im using baseTokenIndex in here few lines lower we have this => 

```
const payerTokenIndex = ((): TokenIndex => {
      if (side == Serum3Side.bid) {
        return serum3Market.quoteTokenIndex;
      } else {
        return serum3Market.baseTokenIndex;
      }
    })();
```

should i use only basetokenIndex or take the index form this function ?

ts/client/src/client.ts
```diff
@@ -2515,6 +2598,12 @@ export class MangoClient {
     // but user would potentially open new positions.
     banks: Bank[],
     perpMarkets: PerpMarket[],
+    additionalAccounts: AdditionalHealthAccounts = {
```
abrzezinski94, line 2601: so because remaining accounts need to be in the right order i added option in here to pass it optionally to build function

ts/client/src/client.ts
```diff
@@ -1258,15 +1336,18 @@ export class MangoClient {
...
-    );
+    txes.push(...placeOrderIxes);
+    if (mangoAccount.getSerum3Account(serum3Market.marketIndex)?.openOrders) {
+      const ix2 = await this.serum3SettleFundsIx(
+        group,
+        mangoAccount,
+        externalMarketPk,
+      );
+      txes.push(ix2);
+    }
```
tylersssss, line 1347: Shouldn't we always do the settle funds ix after placing an order? What is this `if` check for?

abrzezinski94: fixed

ts/client/src/client.ts
abrzezinski94: fixed

ts/client/src/client.ts
```diff
@@ -1141,18 +1193,51 @@ export class MangoClient {
...
+
+      const openOrderPublicKey = await this.findOpenOrdersProgramAddress(
+        mangoAccount.publicKey,
+        serum3Market.publicKey,
+      );
+      const tokenIndex =
+        serum3Market[
+          side == Serum3Side.bid ? 'baseTokenIndex' : 'quoteTokenIndex'
+        ];
+      const baseBank = group.getFirstBankByTokenIndex(tokenIndex);
```
abrzezinski94, line 1218: getFirstBank is the right fcn here ?

PR #368, opened 2023-01-09
Perp: liq with token instruction

ckamm:


programs/mango-v4/src/instructions/perp_liq_with_token.rs
```diff
@@ -0,0 +1,365 @@
...
+    let mut liqee = ctx.accounts.liqee.load_full_mut()?;
+    let mut liqor = ctx.accounts.liqor.load_full_mut()?;
+    // account constraint #1
+    require!(
+        liqor
+            .fixed
+            .is_owner_or_delegate(ctx.accounts.liqor_owner.key()),
+        MangoError::SomeError
+    );
+    require!(!liqor.fixed.being_liquidated(), MangoError::BeingLiquidated);
```
microwavedcola1, line 96: nit: maybe we want something like ::LiqorCannotBeBeingLiquidated

programs/mango-v4/src/health/cache.rs
```diff
@@ -446,13 +448,94 @@ impl HealthCache {
...
+    }
+
+    pub fn has_perp_base_positions(&self) -> bool {
+        self.perp_infos.iter().any(|p| p.base_lots != 0)
+    }
+
+    pub fn has_perp_positive_trusted_pnl(&self) -> bool {
+        self.perp_infos
+            .iter()
+            .any(|p| p.trusted_market && p.quote > 0)
```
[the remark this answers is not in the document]
microwavedcola1: ah this is only called in phase2, where there are no more perp base positions !!

programs/mango-v4/src/instructions/perp_liq_with_token.rs
```diff
@@ -0,0 +1,365 @@
...
+            .max(I80F48::ZERO);
+
+        // Available insurance fund coverage
+        let insurance_vault_amount = if perp_market.elligible_for_group_insurance_fund() {
+            ctx.accounts.insurance_vault.amount
+        } else {
+            0
+        };
+
+        let liquidation_fee_factor = cm!(I80F48::ONE + perp_market.liquidation_fee);
```
microwavedcola1, line 240: is liquidation_fee required, if insurance fund is usdc?

nicholasgodfreyclarke (commented):
Looks good to me. However, should there be a liquidation fee for perp_liq_with_token in the case that liqor exchanges quote position for tokens? (Before the bankruptcy logic).

programs/mango-v4/src/health/cache.rs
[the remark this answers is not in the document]
ckamm: I made this clearer anyway

programs/mango-v4/src/instructions/perp_liq_with_token.rs
microwavedcola1: yeah, but its not a biggie I guess

microwavedcola1: I think a meaningful error msg does save reading code

programs/mango-v4/src/instructions/perp_liq_with_token.rs
ckamm: It's the liqor incentive to make sure they'll want to call it. It may not be strictly needed - similarly to the settlement part - because liqors should be happy to take negative perp pnl and token assets: it doesn't come with risk and allows them to collect interest until it's settled.

microwavedcola1 (approved):

PR #370, opened 2023-01-11
group halt

microwavedcola1:


programs/mango-v4/src/instructions/compute_account_data.rs
```diff
@@ -3,6 +3,9 @@ use anchor_lang::prelude::*;
 
 #[derive(Accounts)]
 pub struct ComputeAccountData<'info> {
+    #[account(
+        constraint = group.load()?.is_operational()
+    )]
```
ckamm, line 8: let's not disable this one, since this is a pure info-gathering instruction

programs/mango-v4/src/instructions/group_halt.rs
```diff
@@ -0,0 +1,19 @@
+use anchor_lang::prelude::*;
+
+use crate::state::*;
+
+#[derive(Accounts)]
+pub struct GroupToggleHalt<'info> {
```
ckamm, line 6: it's called GroupToggleHalt here but the file is group_halt. Either is fine with me

programs/mango-v4/src/instructions/group_edit.rs
```diff
@@ -7,6 +7,7 @@ pub struct GroupEdit<'info> {
     #[account(
         mut,
         has_one = admin,
+        constraint = group.load()?.is_operational()
```
ckamm, line 10: I'd not do the check on the _edit instructions. Maybe the group should be halted and can be restarted once a bunch of edits have been made.

ckamm:
Generally: I think anchor allows tagging constraints with an error type `constraint = x @ MyError` or something. This could be good to use for these constraint checks.

programs/mango-v4/src/instructions/group_edit.rs
microwavedcola1: makes sense, I have removed the constraint in 

group_edit.rs
perp_edit_market.rs
serum3_edit_market.rs
token_edit.rs

ckamm (approved):

PR #371, opened 2023-01-11
Fix ids api

microwavedcola1:


ts/client/src/accounts/group.ts
```diff
@@ -308,7 +310,7 @@ export class Group {
       await client.program.provider.connection.getMultipleAccountsInfo(oracles);
 
     const coder = new BorshAccountsCoder(client.program.idl);
-    ais.forEach(async (ai, i) => {
+    for await (const [i, ai] of ais.entries()) {
```
microwavedcola1, line 313: the await here is key

ts/client/src/accounts/group.ts
```diff
@@ -90,7 +90,9 @@ export class Group {
     public vaultAmountsMap: Map<string, BN>,
   ) {}
 
-  public async reloadAll(client: MangoClient, ids?: Id): Promise<void> {
+  public async reloadAll(client: MangoClient): Promise<void> {
```
microwavedcola1, line 93: removing the ids option, not all pathways were using ids json before, now once configured on the client during building the client, the gpa vs api is used implicitly

ts/client/src/accounts/group.ts
tylersssss: We may be able to make this faster by using Promise.all

tylersssss: something like this:

```
const loadPrices = Array.from(ais.entries()).map(async ([i, ai]) => {
      const perpMarket = perpMarkets[i];
      if (!ai)
        throw new Error(
          `Undefined ai object in reloadPerpMarketOraclePrices for ${perpMarket.oracle}!`,
        );
      const { price, uiPrice } = await this.decodePriceFromOracleAi(
        coder,
        perpMarket.oracle,
        ai,
        perpMarket.baseDecimals,
        client,
      );
      perpMarket._price = price;
      perpMarket._uiPrice = uiPrice;
    });

    await Promise.all(loadPrices);
    ```

microwavedcola1: thanks, fixed

tylersssss (approved):

PR #372, opened 2023-01-12
mango account freeze

microwavedcola1:


ckamm (commented):
We also need to prohibit using a frozen account as `liqor` in any of the liq instructions.

I think it's worthwhile to also (optionally?) prohibit being liquidated or settled.

ckamm (approved):

PR #376, opened 2023-01-12
output to security tab

silas-x:
This PR changes outputs from dependency and TS code scanning tools from the runner to the security tab
image

Makes it easier to track non-critical vulnerabilities and enables us to easily ignore etc. with audit trail.


riordanp (approved):
Nice!

PR #379, opened 2023-01-12
make untrusted till we have figured out perp settling

microwavedcola1:

### DIFF
--- a/ts/client/src/scripts/mb-admin.ts
+++ b/ts/client/src/scripts/mb-admin.ts
@@ -509,6 +509,45 @@ async function makePerpMarketReduceOnly() {
   );
 }
 
+async function makePerpMarketUntrusted() {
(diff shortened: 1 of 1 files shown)


ts/client/src/scripts/mb-admin.ts
```diff
@@ -509,6 +509,45 @@ async function makePerpMarketReduceOnly() {
...
+    false,
+    null,
+    null,
+    null,
+    null,
+    null,
+    null,
+    null,
+    null,
+    null,
```
ckamm, line 547: This is hard to review. I know it's not a priority, but this horrible API really sticks out to me every time I see it. Might be worthwhile to spend some time to change this to something where unchanged parameters don't need to be mentioned.

Separate PR though.

ckamm (approved):

PR #390, opened 2023-01-14
Dont include equity for untrusted perp markets

tylersssss:
This change filters out positive unsettled pnl for untrusted perp markets in a PerpPosition's getEquity calculation. This change was a result of Max's report that his account leverage number was wrong. The leverage calculation in the UI is essentially `mangoAccount.getAssetsValue() / mangoAccount.getEquity()`. The mangoAccount.getEquity() function calls the perpPosition.getEquity() function changed in this PR.

> so i have this account:
> 1) it has $2.09 in unsettled mngo-perp pnl
> 2) it's max long USDC/DAI - 0.029552 / -0.02686591
> 
> this is surprising:
> 1) i would expect the max leverage for a USDC/DAI position to be around 10x on $0.003  but it displays 1x
> 2) my health is at 4% which suggest that i am already leveraged quite high
> 
> i think the issue here is that the leverage calculation does include the perp market somehow in my pnl

image
Screen Shot 2023-01-14 at 3 49 19 PM

There are two possible side effects of this change that I am uncertain of. 

The first is the effect that this will have on `market-maker.ts`. There is code in that file on line 421 that calls `mangoAccount.getEquity()` and mangoAccount.getEquity calls the perp position's `getEquity()` function that is changed in this PR.

The second side effect is from `mangoAccount.getPnl()` which calls mangoAccount.getEquity() which is affected by this change. Though I couldn't find any calls to mangoAccount.getPnl in the client or the UI.

After this change the UI for Max's test account will display the leverage number he expected:

Screen Shot 2023-01-14 at 3 58 21 PM

### DIFF
--- a/ts/client/src/accounts/mangoAccount.ts
+++ b/ts/client/src/accounts/mangoAccount.ts
@@ -1303,7 +1303,13 @@ export class PerpPosition {
       .sub(unsettledFunding)
       .add(takerQuote);
 
-    return baseLots.mul(lotsToQuote).add(quoteCurrent);
+    const uncappedEquity = baseLots.mul(lotsToQuote).add(quoteCurrent);
+
+    if (perpMarket.trustedMarket) {
+      return uncappedEquity;
+    } else {
+      return baseLots.mul(lotsToQuote).add(quoteCurrent).min(ZERO_I80F48());
+    }
   }
 
   public hasOpenOrders(): boolean {


ckamm:
Yes, unsettled pnl in untrusted markets doesn't count to health and likely shouldn't be considered equity. It is pnl though. I'll check in more detail tomorrow.

ckamm:
If we want equity to mean "the USD value you'd get if you settle and sell all your account positions at current oracle prices", then unsettled pnl should contine being added to it. It doesn't take risk into account elsewhere, so also shouldn't worry about the risk of unsettled pnl.

Furthermore, soon there'll no longer be a binary trusted vs untrusted flag for perp markets: instead there'll be numeric asset weights for perp pnl. Would we not count perp pnl that has 80% asset weight for equity, but count token balances that have 80% weight?

tylersssss:
image

ckamm:
I think `getAssetsValue()` should include unsettled perp pnl. That way it returns the quote value the user would receive if he could settle/sell all assets at the current price.

For consistency we should always have `getEquity() = getAssetsValue() - getLiabsValue()`. 

That means that in an example with 0.03 spot assets, -0.025 spot liabs and 2 unsettled pnl, the leverage would be very low:
```
leverage = assets_value / equity_value - 1 = 2.03 / 2.005 - 1 = 0.012x
```
Which makes sense imo because that's also what it would be once the perp pnl is settled.

Yes, health is very low at the same time. But that's just because leverage defined this way is not always correlated with health in v4.

There are other cases where leverage and health would diverge. For example, if a user deposited $100M mSOL but we had set a $1M soft deposit limit, then the user could probably borrow up to around $800k USDC. Then their init health would be 0. But their leverage defined this way would just be `100M / 99.2M -1 = 0.008x.

ckamm:
@tylersssss this can be closed, right?

tylersssss:
> @tylersssss this can be closed, right?
yep, thanks

PR #391, opened 2023-01-16
Perp: Allow setting pnl asset weights

ckamm:
This replaces the previous distinction between trusted and untrusted markets, they are equivalent to setting the asset weights to 1 or 0 instead.

This way, we can weigh positive pnl in the trusted case at less than 1 which is more correct from a risk point of view and allows for more flexibility when it comes to liquidation.


microwavedcola1 (approved):

ts/client/src/scripts/mb-admin.ts
```diff
@@ -424,6 +424,8 @@ async function registerPerpMarkets() {
     0.95,
     1.025,
     1.05,
+    1,
```
microwavedcola1, line 427: we probably want to edit btc-perp and set some new maintPnlAssetWeight and initPnlAssetWeight

microwavedcola1: also what should be the defaults for this for previously trusted markets?

ckamm: Don't change BTC-PERP yet, we still need to add an instruction for the liquidation of positive pnl that can now take advantage of the difference between the asset weight and 1 to give a liquidation incentive to the liqor.

Defaults: It expresses the risk from unsettled pnl potentially never getting settled. Maybe something like 0.9 / 0.95 for markets we're highly confident in? We need to talk about it.

PR #392, opened 2023-01-16
Perps: track overall realized pnl relating to a position

ckamm:
This includes trade pnl, funding and fees. Tracking this makes it easier for uis to display a consistent position overall pnl value that doesn't decrease by settling.

### DIFF
--- a/ts/client/src/accounts/mangoAccount.ts
+++ b/ts/client/src/accounts/mangoAccount.ts
@@ -175,7 +175,6 @@ export class MangoAccount {
 
   public getPerpPosition(
     perpMarketIndex: PerpMarketIndex,
-    useEventQueue?: boolean,
   ): PerpPosition | undefined {
     return this.perps.find((pp) => pp.marketIndex == perpMarketIndex);
   }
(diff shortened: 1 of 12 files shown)


microwavedcola1 (approved):

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1318,13 +1322,13 @@ export class PerpPosition {
     );
   }
 
-  public getAverageEntryPriceUi(perpMarket: PerpMarket): number {
-    if (perpMarket.perpMarketIndex !== this.marketIndex) {
-      throw new Error("PerpPosition doesn't belong to the given market!");
-    }
+  public getAverageEntryPrice(perpMarket: PerpMarket): I80F48 {
+    return I80F48.fromI64(this.basePositionLots.mul(perpMarket.baseLotSize));
```
ckamm, line 1326: this isn't the average entry price!

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1342,15 +1346,40 @@ export class PerpPosition {
...
     }
 
-    return this.quotePositionNative.add(
-      I80F48.fromI64(this.basePositionLots.mul(perpMarket.baseLotSize)).mul(
-        perpMarket.price,
+    const priceChange = perpMarket.price.sub(
+      this.getAverageEntryPrice(perpMarket),
+    );
+
+    return toUiDecimals(
```
ckamm, line 1361: could be worth to add a `perpMarket.settleTokenNativeToUi()`

microwavedcola1: I will do this later, its a lightly larger change I realise now

PR #394, opened 2023-01-17
security admin can bring markets to reduce only

microwavedcola1:


programs/mango-v4/src/instructions/perp_edit_market.rs
```diff
@@ -52,6 +49,48 @@ pub fn perp_edit_market(
...
+        && settle_fee_flat_opt.is_none()
+        && settle_fee_amount_threshold_opt.is_none()
+        && settle_fee_fraction_low_health_opt.is_none()
+        && stable_price_delay_interval_seconds_opt.is_none()
+        && stable_price_delay_growth_limit_opt.is_none()
+        && stable_price_growth_limit_opt.is_none()
+        && settle_pnl_limit_factor_opt.is_none()
+        && settle_pnl_limit_window_size_ts_opt.is_none()
+        // security admin can bring to reduce only mode
+        && reduce_only_opt.is_some()
```
ckamm, line 80: I'm concerned about code like this because someone will add a new field to perp_edit_market and forget to list it here.

Alternate suggestion:
```
let mut require_group_admin = false;

if let ... {
   ...
   require_group_admin = true; // do this on every single change that needs the group admin
}

// lots more seting of fields

// at the end
if require_group_admin {
   require!(...)
}
```

programs/mango-v4/src/instructions/perp_edit_market.rs
```diff
@@ -74,112 +70,128 @@ pub fn perp_edit_market(
...
+        require_group_admin = true;
     }
     if let Some(settle_pnl_limit_window_size_ts) = settle_pnl_limit_window_size_ts_opt {
         perp_market.settle_pnl_limit_window_size_ts = settle_pnl_limit_window_size_ts;
+        require_group_admin = true;
     }
 
     if let Some(reduce_only) = reduce_only_opt {
         perp_market.reduce_only = u8::from(reduce_only);
     };
```
ckamm, line 179: We should only allow the security authority to switch on reduce only, not to switch it off.

programs/mango-v4/src/instructions/perp_edit_market.rs
```diff
@@ -74,112 +70,138 @@ pub fn perp_edit_market(
...
-
     if let Some(maint_base_asset_weight) = maint_base_asset_weight_opt {
         perp_market.maint_base_asset_weight = I80F48::from_num(maint_base_asset_weight);
+        require_group_admin = true;
     }
     if let Some(init_base_asset_weight) = init_base_asset_weight_opt {
+        let old_init_base_asset_weight = perp_market.init_base_asset_weight;
         perp_market.init_base_asset_weight = I80F48::from_num(init_base_asset_weight);
+
+        // security admin can only reduce init_base_asset_weight
```
ckamm, line 84: for untrusted operation we also need to check >= 0, sorry only thought of that just now

programs/mango-v4/src/instructions/perp_edit_market.rs
```diff
@@ -74,112 +70,138 @@ pub fn perp_edit_market(
...
     if let Some(settle_pnl_limit_window_size_ts) = settle_pnl_limit_window_size_ts_opt {
         perp_market.settle_pnl_limit_window_size_ts = settle_pnl_limit_window_size_ts;
+        require_group_admin = true;
     }
 
     if let Some(reduce_only) = reduce_only_opt {
         perp_market.reduce_only = u8::from(reduce_only);
+
+        // security admin can only enable reduce_only
+        if perp_market.reduce_only == u8::from(false) {
```
ckamm, line 186: nit: would be easier to say `if !reduce_only {`

programs/mango-v4/src/instructions/perp_edit_market.rs
microwavedcola1: wdym by that? so something like

image

ckamm: yeah, this or just generally requiring >= 0. Negative weights don't make sense. We can trust the admin, but not the security authority.

microwavedcola1: I think I will apply it generally

ckamm (approved):

PR #395, opened 2023-01-17
Add oracleLastUpdatedSlot to Bank and PerpMarket

riordanp:
Parses the best value we have for last updated slot and sets it on the Bank and PerpMarkets when reloading oracle prices. Used for checking oracle staleness on the client side.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "validate": "npm run typecheck && npm run lint && npm run format"
   },
   "devDependencies": {
-    "@jup-ag/core": "^1.0.0-beta.28",
     "@tsconfig/recommended": "^1.0.1",
     "@types/bs58": "^4.0.1",
     "@types/chai": "^4.3.0",
@@ -58,6 +57,7 @@
     "trailingComma": "all"
   },
   "dependencies": {
+    "@jup-ag/core": "^1.0.0-beta.28",
     "@project-serum/anchor": "^0.25.0",
     "@project-serum/serum": "^0.13.65",
     "@pythnetwork/client": "~2.8.0",
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
     "trailingComma": "all"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


ts/client/src/accounts/oracle.ts
```diff
@@ -48,21 +48,29 @@ export class StubOracle {
...
   accountInfo: AccountInfo<Buffer>,
-): number {
-  const aggregatorAccountData =
-    program.decodeLatestAggregatorValue(accountInfo);
-  if (!aggregatorAccountData)
+): { price: number; lastUpdatedSlot: number } {
+  const price = program.decodeLatestAggregatorValue(accountInfo)!.toNumber();
+  const lastUpdatedSlot = program
+    .decodeAggregator(accountInfo)
+    .previousConfirmedRoundSlot!.toNumber();
```
microwavedcola1, line 69: hmm should this rather be latest_confirmed_round? @ckamm

riordanp: yes this matches the check in the program, updated

package.json
```diff
@@ -58,6 +57,7 @@
     "trailingComma": "all"
   },
   "dependencies": {
+    "@jup-ag/core": "^1.0.0-beta.28",
```
tylersssss, line 60: Is this package needed? I don't see it imported anywhere in this PR or in my local copy. 

I think it was added a long time ago and can now be deleted.

tylersssss, line 60: Is this package needed? I don't see it imported anywhere in this PR or in my local copy. 

I think it was added a long time ago and can now be deleted.

tylersssss (changes requested):

package.json
tylersssss: Not sure how I ended up with 2 of the same comment. Resolving this one.

riordanp: Seems to build ok without it

microwavedcola1 (approved):

tylersssss (approved):

PR #396, opened 2023-01-17
getAverageEntryPrice should div instead of mul

tylersssss:
I think getAverageEntryPrice should div instead of mul. This matches the previous code before latest changes and the numbers look correct in the UI when dividing.


ckamm (changes requested):
The change to `div` is correct, but this PR has a bunch of unrelated commits, including a merge

PR #403, opened 2023-01-18
fix bug where serum oo where not loaded

microwavedcola1:
Serum OO accounts are needed to do a full health computation. But these are only loaded via mango account reload or reloadAccountData methods. Now this side effect is already achievable while fetching the mango account with an explicity flag. Also reloadAccountData is renamed to be more specific, since we dont really load any other data

### DIFF
--- a/ts/client/src/accounts/perp.ts
+++ b/ts/client/src/accounts/perp.ts
@@ -431,7 +431,17 @@ export class PerpMarket {
     direction: 'negative' | 'positive',
     count = 2,
   ): Promise<{ account: MangoAccount; settleablePnl: I80F48 }[]> {
-    let accountsWithSettleablePnl = (await client.getAllMangoAccounts(group))
+    let accountsWithSettleablePnl = (
+      await Promise.all(
+        await (
+          await client.getAllMangoAccounts(group)
+        ).map(async (a) => {
+          // Need to load serum oo accounts to compute settle perp health
+          await a.reloadAccountData(client);
+          return a;
+        }),
+      )
+    )
       .filter((acc) => acc.perpPositionExistsForMarket(this))
       .map((acc) => {
         const pp = acc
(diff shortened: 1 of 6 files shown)


ts/client/src/accounts/perp.ts
```diff
@@ -431,7 +431,9 @@ export class PerpMarket {
     direction: 'negative' | 'positive',
     count = 2,
   ): Promise<{ account: MangoAccount; settleablePnl: I80F48 }[]> {
-    let accountsWithSettleablePnl = (await client.getAllMangoAccounts(group))
+    let accountsWithSettleablePnl = (
+      await await client.getAllMangoAccounts(group, true)
```
microwavedcola1, line 435: this was where the issue manifested

ts/client/src/client.ts
```diff
@@ -668,28 +678,39 @@ export class MangoClient {
...
       mangoAccountPk,
       await this.program.account.mangoAccount.fetch(mangoAccountPk),
     );
+    if (loadSerum3Oo) {
+      await mangoAccount_?.reloadSerum3OpenOrders(this);
+    }
+    return mangoAccount_;
   }
 
   public async getMangoAccountForPublicKey(
```
ckamm, line 697: what's the difference between getMangoAccount and getMangoAccountForPublicKey -- doesn't getMangoAccount cover the latter function?

ts/client/src/client.ts
```diff
@@ -701,27 +722,36 @@ export class MangoClient {
...
+    loadSerum3Oo = false,
   ): Promise<MangoAccount | undefined> {
     const mangoAccounts = await this.getMangoAccountsForOwner(group, ownerPk);
     const foundMangoAccount = mangoAccounts.find(
       (a) => a.accountNum == accountNumber,
     );
 
+    if (loadSerum3Oo) {
+      await foundMangoAccount?.reloadSerum3OpenOrders(this);
+    }
```
ckamm, line 744: pass the flag on to getMangoAccountsForOwner, don't need to do this here then

PR #404, opened 2023-01-19
dont process orders which have 0 lots; dont return early, we still want to apply ioc penalty

microwavedcola1:

### DIFF
--- a/programs/mango-v4/src/state/orderbook/book.rs
+++ b/programs/mango-v4/src/state/orderbook/book.rs
@@ -155,8 +155,6 @@ impl<'a> Orderbook<'a> {
             let match_base_lots = remaining_base_lots
                 .min(best_opposing.node.quantity)
                 .min(max_match_by_quote);
-            let done =
-                match_base_lots == max_match_by_quote || match_base_lots == remaining_base_lots;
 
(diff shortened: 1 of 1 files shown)


programs/mango-v4/src/state/orderbook/book.rs
```diff
@@ -114,6 +114,10 @@ impl<'a> Orderbook<'a> {
         let opposing_bookside = self.bookside_mut(other_side);
         for best_opposing in opposing_bookside.iter_all_including_invalid(now_ts, oracle_price_lots)
         {
+            if remaining_base_lots == 0 || remaining_quote_lots == 0 {
```
microwavedcola1, line 117: either I can do this, or the even less invasive change would be to only skip the insertion of 0 match_base_lots even to event queue part

ckamm: This exist condition is good. With this, we can also remove the `if done { break }` and the whole `done` variable later in the function.

microwavedcola1: done

ckamm (approved):

PR #409, opened 2023-01-20
ensure that freshly fetched objects have valid prices until we fetch …

microwavedcola1:
…oracles again

Currently group.reloadAll does this - it fetches banks, and then sets prices on banks by refetching oracles, for a brief period, between having fetched new banks and having to fetch oracles, the prices on banks are undefined, lets just  persist with old values until new are present 

### DIFF
--- a/ts/client/src/accounts/group.ts
+++ b/ts/client/src/accounts/group.ts
@@ -8,6 +8,7 @@ import {
   PublicKey,
 } from '@solana/web3.js';
 import BN from 'bn.js';
+import _ from 'lodash';
 import { MangoClient } from '../client';
 import { OPENBOOK_PROGRAM_ID } from '../constants';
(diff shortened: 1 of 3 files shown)


ts/client/src/accounts/group.ts
```diff
@@ -8,6 +8,7 @@ import {
   PublicKey,
 } from '@solana/web3.js';
 import BN from 'bn.js';
+import _ from 'lodash';
```
tylersssss, line 11: we should import only the functions we need here so that we get the benefits of tree shaking and a smaller bundle size

`import { cloneDeep, merge } from 'lodash';`

tylersssss (changes requested):

tylersssss (approved):

PR #411, opened 2023-01-20
abort if new tokens or perp markets listed

microwavedcola1:

### DIFF
--- a/keeper/src/crank.rs
+++ b/keeper/src/crank.rs
@@ -112,15 +112,6 @@ pub async fn loop_update_index_and_rate(
 ) {
     let mut interval = time::interval(Duration::from_secs(interval));
     loop {
-        if mango_client
-            .context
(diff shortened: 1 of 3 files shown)


ckamm (commented):
Basically this works, but the frequency of checking for new tokens/perp markets seems high to me. How about starting a separate job that checks these >= every minute or so?

keeper/src/crank.rs
```diff
@@ -73,12 +74,37 @@ pub async fn runner(
...
+            mango_client.clone(),
+            interval_check_new_listings_and_abort
+        ),
         debugging_handle
     );
 
     Ok(())
 }
 
+pub async fn loop_check_new_listings_and_abort(mango_client: Arc<MangoClient>, interval: u64) {
```
ckamm, line 87: general thought, not for this PR: we should use rust's Duration type for all of these

ckamm (approved):

PR #412, opened 2023-01-20
Perp: Clarify has_open_orders use during liquidation

ckamm:


liquidator/src/liquidate.rs
```diff
@@ -168,7 +168,7 @@ impl<'a> LiquidateHelper<'a> {
         > = stream::iter(self.liqee.active_perp_positions())
             .then(|pp| async {
                 let base_lots = pp.base_position_lots();
-                if base_lots == 0 {
+                if base_lots == 0 || pp.has_open_taker_fills() {
```
microwavedcola1, line 171: so I am guessing thought here is that the user reduced his position, and that means liqor would take on even less liabs? He obviously can't increase if he is liquidatable

liquidator/src/liquidate.rs
```diff
@@ -572,18 +572,26 @@ impl<'a> LiquidateHelper<'a> {
...
         if let Some(txsig) = self.perp_close_orders().await? {
-            return Ok(txsig);
+            return Ok(Some(txsig));
         }
         if let Some(txsig) = self.serum3_close_orders().await? {
-            return Ok(txsig);
+            return Ok(Some(txsig));
+        }
+
+        if self.health_cache.has_phase1_liquidatable() {
```
microwavedcola1, line 581: is the local copy of account updated between cancelling perp/serum orders and here? if not this would always return true no?

liquidator/src/liquidate.rs
```diff
@@ -572,18 +572,26 @@ impl<'a> LiquidateHelper<'a> {
         //
         // TODO: All these close ix could be in one transaction.
         if let Some(txsig) = self.perp_close_orders().await? {
-            return Ok(txsig);
+            return Ok(Some(txsig));
         }
         if let Some(txsig) = self.serum3_close_orders().await? {
```
microwavedcola1, line 577: this cancels orders on a random market no? do you want to add todo to cancel all in future before moving on, or is that by design?

liquidator/src/liquidate.rs
```diff
@@ -596,7 +604,24 @@ impl<'a> LiquidateHelper<'a> {
...
+        if self.health_cache.has_perp_open_fills() {
+            log::info!(
+                "Account {} has open perp fills, maint_health {}, waiting...",
+                self.pubkey,
+                self.maint_health
+            );
+            return Ok(None);
+        }
+
+        if self.health_cache.has_phase2_liquidatable() {
```
microwavedcola1, line 619: technically this check would cover the one on 610, but ok the one on 610 comes with an explicit msg about open perp fills

microwavedcola1 (approved):

liquidator/src/liquidate.rs
ckamm: Yeah, the thought is that reducing base_position to 0 is useless when the incoming fill is going to change it anyway. Possibly we could aim for bringing the effective_base_pos to 0, but it may be safer to just wait for the fill to happen - that will also change the user health, after all.

liquidator/src/liquidate.rs
ckamm: This currently runs one step of the liquidation and then returns. So yeah, the next time this function is called, it'll work from an updated account state.

liquidator/src/liquidate.rs
ckamm: It cancels a random one each time. Next time it'll cancel the next. There's a TODO to just cancel all in one tx in there somewhere.

liquidator/src/liquidate.rs
ckamm: The key difference is that 610 is info-log-and-wait, while this is log-critical-error

PR #415, opened 2023-01-23
group level token deposit limit

microwavedcola1:

### DIFF
--- a/programs/mango-v4/src/instructions/token_deposit.rs
+++ b/programs/mango-v4/src/instructions/token_deposit.rs
@@ -189,11 +189,13 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
         // Since depositing can only increase health, we can skip the usual pre-health computation.
         // Also, TokenDeposit is one of the rare instructions that is allowed even during being_liquidated.
         //
-        if !account.fixed.is_in_health_region() {
+        // TEMPORARY: check is disabled, since we want to enforce a temporary deposit limit
+        // if !account.fixed.is_in_health_region()
+        {
             let retriever =
                 new_fixed_order_account_retriever(remaining_accounts, &account.borrow())?;
-            let health = compute_health(&account.borrow(), HealthType::Init, &retriever)
-                .context("post-deposit init health")?;
+            let cache = new_health_cache(&account.borrow(), &retriever)?;
+            let health = cache.health(HealthType::Init);
             msg!("health: {}", health);
             let was_being_liquidated = account.being_liquidated();
             let recovered = account.fixed.maybe_recover_from_being_liquidated(health);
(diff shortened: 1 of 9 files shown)


programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -189,18 +189,38 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
...
-                .context("post-deposit init health")?;
+            let cache = new_health_cache(&account.borrow(), &retriever)?;
+            let health = cache.health(HealthType::Init);
             msg!("health: {}", health);
             let was_being_liquidated = account.being_liquidated();
             let recovered = account.fixed.maybe_recover_from_being_liquidated(health);
             require!(
                 !was_being_liquidated || recovered,
                 MangoError::DepositsIntoLiquidatingMustRecover
             );
```
ckamm, line 206: Let's make this not a temporary hack. token_deposit would always build the retriever and cache, but this health check etc would be gated by `!is_in_health_region()`.

programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -189,18 +189,38 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
...
+                .unwrap();
+            let group = self.group.load()?;
+            if assets > group.deposit_limit_quote {
+                return err!(MangoError::DepositLimit).with_context(|| {
+                    format!(
+                        "assets ({:?}) can't cross deposit limit on the group ({:?})",
+                        assets, group.deposit_limit_quote,
+                    )
+                });
+            }
```
ckamm, line 223: this whole thing should be behing a `if deposit_limit_quote > 0`, because we want the default state to be "no deposit limit", I think

but then we can't support a literal 0 deposit limit. We could either store an Option<u64> on the group, or just be ok with it (needs docs on the field though!) and say 1e-6 USD is the minimum deposit limit

programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -189,18 +189,38 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
...
+                .0
+                .round_to_zero()
+                .checked_to_num::<u64>()
+                .unwrap();
+            let group = self.group.load()?;
+            if assets > group.deposit_limit_quote {
+                return err!(MangoError::DepositLimit).with_context(|| {
+                    format!(
+                        "assets ({:?}) can't cross deposit limit on the group ({:?})",
+                        assets, group.deposit_limit_quote,
```
ckamm, line 220: use
```
require_msg_typed!(
    assets <= group.deposit_limit_quote, 
    MangoError::DepositLimit,
    "assets ({}) can't cross deposit limit on the group ({})",
    assets, 
    group.deposit_limit_quote,

programs/mango-v4/src/instructions/token_deposit.rs
microwavedcola1: so what I have done is - I have made group.deposit_limit_quote a signed integer now, and setting to -1 basically disables it, this allows us to set any non-negative number as deposit limit including zero, wdyt?

ckamm: That means groups have a 0 deposit limit by default, and once we push the program update, deposits would be disabled until we group_edit (and we need to do this on all existing groups).

If you want to go that way, group_create should be updated to let it start out at -1, otherwise all test scripts will break too. It's probably better to let the zero state be "no deposit limit"?

microwavedcola1: yeah I was working on seting defaults in group create already, but ok, I will set 0 no deposit limit, and everything higher to an actual enforcement

microwavedcola1: done

ckamm (approved):

PR #417, opened 2023-01-23
return order id from a perp place order ix

microwavedcola1:

### DIFF
--- a/ts/client/src/client.ts
+++ b/ts/client/src/client.ts
@@ -1919,7 +1919,7 @@ export class MangoClient {
       .rpc();
   }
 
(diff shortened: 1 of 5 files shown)


ts/client/src/client.ts
```diff
@@ -1919,6 +1919,9 @@ export class MangoClient {
       .rpc();
   }
 
+  // perpPlaceOrder ix returns an optioanl order id,
+  // but, since we use a customer tx sender, this method
```
ckamm, line 1923: optional, custom

ckamm (approved):

PR #423, opened 2023-01-24
logging + prevention of setting admin to default pk

microwavedcola1:


ckamm (approved):

programs/mango-v4/src/instructions/group_edit.rs
```diff
@@ -26,26 +26,45 @@ pub fn group_edit(
     let mut group = ctx.accounts.group.load_mut()?;
 
     if let Some(admin) = admin_opt {
+        require_keys_neq!(admin, Pubkey::default(), MangoError::SomeError);
```
ckamm, line 29: nit: I don't think you need to provide an error type to require_keys_neq!() unless you really want to

PR #424, opened 2023-01-24
fix liquidating perp base positions with 0 overall perp weight

ckamm:


programs/mango-v4/src/instructions/perp_liq_base_position.rs
```diff
@@ -96,41 +128,42 @@ pub fn perp_liq_base_position(
...
     // Settle funding
     liqee_perp_position.settle_funding(&perp_market);
     liqor_perp_position.settle_funding(&perp_market);
 
+    // Max settleable on the liqee?
+    let now_ts: u64 = Clock::get()?.unix_timestamp.try_into().unwrap();
+    liqee_perp_position.update_settle_limit(&perp_market, now_ts);
+    let liqee_positive_settle_limit = liqee_perp_position.available_settle_limit(&perp_market).1;
+    let max_settle = I80F48::from(max_quote_transfer)
+        .min(liqor_settle_health)
```
microwavedcola1, line 141: note to self: not obvious why liqor_settle_health should be the lower constraint

microwavedcola1, line 141: note to self: not obvious why this is used here

ckamm: I think it's not strictly necessary in this case.

For normal settlement it's essential because settlement is health-neutral on the account with negative pnl, but converts negative perp health to negative spot health -- and we don't want to allow converting perp losses into unbacked spot borrows this way.

In this case, the liqor pays `1 - liq_fee` spot for every 1 quote he takes over, so for each unit of perp pnl taken over, his health changes by `overall_perp_asset_weight - (1 - liq_fee) < 0`. That automatically limits the amount a liqor can take on.

programs/mango-v4/src/instructions/perp_liq_base_and_positive_pnl.rs
```diff
@@ -0,0 +1,434 @@
...
+        address = settle_bank.load()?.vault
+    )]
+    pub settle_vault: Account<'info, TokenAccount>,
+
+    /// CHECK: Oracle can have different account types
+    #[account(address = settle_bank.load()?.oracle)]
+    pub settle_oracle: UncheckedAccount<'info>,
+}
+
+pub fn perp_liq_base_and_positive_pnl(
```
microwavedcola1, line 57: nit: small comment why positive pnl and negative pnl are separate ixs

programs/mango-v4/src/instructions/perp_liq_base_and_positive_pnl.rs
```diff
@@ -0,0 +1,434 @@
...
+    liqee_perp_position.settle_funding(&perp_market);
+    liqor_perp_position.settle_funding(&perp_market);
+
+    // Max settleable on the liqee?
+    let now_ts: u64 = Clock::get()?.unix_timestamp.try_into().unwrap();
+    liqee_perp_position.update_settle_limit(&perp_market, now_ts);
+    let liqee_positive_settle_limit = liqee_perp_position.available_settle_limit(&perp_market).1;
+
+    // The max settleable amount does not need to be constrained by the liqor's perp settle health,
+    // because taking over perp pnl decreases liqor health: every unit of pnl taken costs
```
microwavedcola1, line 137: nit: rephrase to  `taking over perp position`,

programs/mango-v4/src/instructions/perp_liq_base_and_positive_pnl.rs
```diff
@@ -0,0 +1,434 @@
...
+        direction = 1;
+        fee_factor = cm!(I80F48::ONE + perp_market.base_liquidation_fee);
+        unweighted_health_per_lot = cm!(price_per_lot
+            * (perp_market.init_base_liab_weight - quote_init_liab_weight * fee_factor));
+    };
+    assert!(unweighted_health_per_lot > 0);
+
+    let spot_gain_per_settled = cm!(I80F48::ONE - perp_market.positive_pnl_liquidation_fee);
+    let init_overall_asset_weight = perp_market.init_pnl_asset_weight;
+    let expected_perp_health = |unweighted: I80F48| {
```
microwavedcola1, line 182: nit: add comment that this is done here, instead of using the cache for cu saving?

programs/mango-v4/src/instructions/perp_liq_base_and_positive_pnl.rs
```diff
@@ -0,0 +1,434 @@
...
+        let quote_init_asset_weight = I80F48::ONE;
+        direction = -1;
+        fee_factor = cm!(I80F48::ONE - perp_market.base_liquidation_fee);
+        unweighted_health_per_lot = cm!(price_per_lot
+            * (-perp_market.init_base_asset_weight + quote_init_asset_weight * fee_factor));
+    } else {
+        // liqee_base_lots <= 0
+        require_msg!(
+            max_base_transfer <= 0,
+            "max_base_transfer can't be positive when liqee's base_position is positive"
```
microwavedcola1, line 167: when liqee's base_position is `negative`

programs/mango-v4/src/instructions/perp_liq_base_and_positive_pnl.rs
```diff
@@ -0,0 +1,434 @@
...
+        require_msg!(
+            max_base_transfer >= 0,
+            "max_base_transfer can't be negative when liqee's base_position is positive"
+        );
+
+        // the unweighted perp health contribution gets reduced by `base * price * perp_init_asset_weight`
+        // and increased by `base * price * (1 - liq_fee) * quote_init_asset_weight`
+        let quote_init_asset_weight = I80F48::ONE;
+        direction = -1;
+        fee_factor = cm!(I80F48::ONE - perp_market.base_liquidation_fee);
```
microwavedcola1, line 160: so would base_liquidation_fee>= positive_pnl_liquidation_fee in practice?

programs/mango-v4/src/instructions/perp_liq_base_and_positive_pnl.rs
```diff
@@ -0,0 +1,434 @@
...
+            &mut current_unweighted_perp_health,
+        );
+    }
+
+    //
+    // Execute the base reduction. This is essentially a forced trade and updates the
+    // liqee and liqors entry and break even prices.
+    //
+    let base_transfer = cm!(direction * base_reduction);
+    let quote_transfer = cm!(-I80F48::from(base_transfer) * price_per_lot * fee_factor);
```
microwavedcola1, line 292: hmm one could aruge that you could have separate base_reductions for each step, and then have different fee_factors for them, but that would be too complicated I guess

programs/mango-v4/src/instructions/perp_liq_base_and_positive_pnl.rs
```diff
@@ -0,0 +1,434 @@
...
+        let health_per_settle = cm!(spot_gain_per_settled - perp_market.init_pnl_asset_weight);
+        let settle_for_zero = cm!(-current_actual_health / health_per_settle)
+            .checked_ceil()
+            .unwrap();
+        let liqee_pnl = liqee_perp_position.unsettled_pnl(&perp_market, oracle_price)?;
+
+        // Allow settling *more* than the liqee_positive_settle_limit. In exchange, the liqor
+        // also can't settle fully immediately and just takes over a fractional chunk of the limit.
+        let settlement = liqee_pnl
+            .min(max_settle)
```
microwavedcola1, line 323: why allow settling more than the settle limit?

programs/mango-v4/src/instructions/perp_liq_base_and_positive_pnl.rs
ckamm: can't say

programs/mango-v4/src/instructions/perp_liq_base_and_positive_pnl.rs
ckamm: I think having one fee is sensible, because it mostly compensates the liqor's risk. The liqor doesn't care what step the base position came from, just that he's compensated for the risk of taking it on.

programs/mango-v4/src/lib.rs
```diff
@@ -545,9 +548,10 @@ pub mod mango_v4 {
             stable_price_delay_growth_limit_opt,
             stable_price_growth_limit_opt,
             settle_pnl_limit_factor_opt,
-            settle_pnl_limit_window_size_ts,
+            settle_pnl_limit_window_size_ts_opt,
```
microwavedcola1, line 551: I think this is missing the corresponding ts client change

programs/mango-v4/tests/cases/test_liq_perps_positive_pnl.rs
```diff
@@ -0,0 +1,416 @@
...
+    // SETUP: Increase the price of the borrow so account_0 becomes liquidatable
+    //
+    set_bank_stub_oracle_price(solana, group, &borrow_token, admin, 10.0).await;
+    assert_eq!(
+        account_init_health(solana, account_0).await.round(),
+        (10000.0f64 - 10.0 * 1000.5 * 1.4).round()
+    );
+
+    //
+    // TEST: Can't liquidate base if health wouldn't go up: no effect
```
microwavedcola1, line 229: Some logging in the ix for this case would be a nice to have

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
```diff
@@ -0,0 +1,825 @@
...
+            settle_bank.deposit(liqee_token_position, token_transfer, now_ts)?;
+            settle_bank.withdraw_without_fee(
+                liqor_token_position,
+                token_transfer,
+                now_ts,
+                oracle_price,
+            )?;
+            liqee_health_cache.adjust_token_balance(&settle_bank, token_transfer)?;
+        }
+        msg!("pnl: {}, quote = {}", settlement, token_transfer);
```
microwavedcola1, line 482: maybe rephrase this to - `pnl {} was transferred to liqor for quote {}`

programs/mango-v4/src/lib.rs
ckamm: The ts client passes:
```
        params.settlePnlLimitWindowSize !== null
          ? new BN(params.settlePnlLimitWindowSize)
          : null,
```
so it looks like it doesn't use the _opt name anyway. Did you mean something else?

programs/mango-v4/tests/cases/test_liq_perps_positive_pnl.rs
ckamm: I've changed the instruction to emit perp balance logs as well as health logs when it doesn't think it can do anything useful

PR #427, opened 2023-02-01
Add explicit self-liquidation checks

ckamm:
Just for safety, the other conditions and borrowing pattern would already make it impossible.


microwavedcola1:
Is there a net benefit if the liqee is able to actually self liquidate? if not, then the checks seem unnecessary

microwavedcola1 (approved):

PR #430, opened 2023-02-02
remove perp zero out, was only added to cleanup testing perp market, …

microwavedcola1:
…can be re-added if needed


programs/mango-v4/src/instructions/perp_zero_out.rs
```diff
@@ -1,47 +0,0 @@
-use anchor_lang::prelude::*;
-
-use crate::error::*;
-use crate::state::*;
-
-#[derive(Accounts)]
-pub struct PerpZeroOutForMarket<'info> {
-    #[account(
-        has_one = admin,
-        constraint = group.load()?.is_testing()
```
microwavedcola1, line 10: note: no ix gating, so nothing to be done

ckamm (approved):

PR #436, opened 2023-02-03
Add mm dockerfile and fly conf

microwavedcola1:


microwavedcola1:
@riordanp  how does one use this from cli? could you paste some useful commands

Dockerfile.mm
```diff
@@ -0,0 +1,31 @@
...
+
+#######################################################################
+
+RUN mkdir /app
+WORKDIR /app
+
+COPY . .
+
+RUN yarn install && yarn run build
+FROM debian:bullseye
```
silas-x, line 20: nitpick: makes it easier to review
```suggestion

FROM debian:bullseye as run
```

PR #437, opened 2023-02-04
update serum to most recent openbook version

mschneider:
i am having issues in solana-accountsdb-connector when using both mango & openbook, relaxing the requirements here simplifies my life


ckamm (approved):
Yay for fewer forked dependencies!

Is it correct to point to openbook-dex master? Probably it's slow moving enough to not matter.

mschneider:
yeah, we don't have another target rn. you still need to run cargo update to actually update the dependency so not a big deal imo even if the master includes changes that are not released yet

PR #442, opened 2023-02-08
configure max staleness slot to 120

microwavedcola1:


programs/mango-v4/src/state/oracle.rs
```diff
@@ -77,7 +77,7 @@ impl OracleConfigParams {
     pub fn to_oracle_config(&self) -> OracleConfig {
         OracleConfig {
             conf_filter: I80F48::from_num(self.conf_filter),
-            max_staleness_slots: self.max_staleness_slots.map(|v| v as i64).unwrap_or(-1),
+            max_staleness_slots: self.max_staleness_slots.map(|v| v as i64).unwrap_or(120),
```
ckamm, line 80: this means you can no longer fully disable oracle staleness; wouldn't do that

PR #445, opened 2023-02-09
update settle limit before returning settleable pnl

microwavedcola1:

### DIFF
--- a/ts/client/src/accounts/mangoAccount.ts
+++ b/ts/client/src/accounts/mangoAccount.ts
@@ -1468,11 +1468,39 @@ export class PerpPosition {
     if (perpMarket.perpMarketIndex !== this.marketIndex) {
       throw new Error("PerpPosition doesn't belong to the given market!");
     }
-
+    this.updateSettleLimit(perpMarket);
     return this.applyPnlSettleLimit(
       this.getUnsettledPnl(perpMarket),
       perpMarket,
     );
+  }
+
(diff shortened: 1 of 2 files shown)


ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1380,6 +1380,8 @@ export class PerpPosition {
       throw new Error("PerpPosition doesn't belong to the given market!");
     }
 
+    this.updateSettleLimit(perpMarket);
+
```
ckamm, line 1384: Mh, this is a side-effect in what looks like a read-only accessor function and it doesn't even change what is returned by it. Why is it good to place it here?

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1464,15 +1464,58 @@ export class PerpPosition {
...
-
-    return this.applyPnlSettleLimit(
+    this.updateSettleLimit(perpMarket);
+    const perpSettleHealth = account.getPerpSettleHealth(group);
+    const limitedUnsettled = this.applyPnlSettleLimit(
       this.getUnsettledPnl(perpMarket),
       perpMarket,
     );
+    if (limitedUnsettled.lt(ZERO_I80F48())) {
+      return limitedUnsettled.max(perpSettleHealth.max(ZERO_I80F48().neg()));
```
ckamm, line 1482: parentheses typo: max(ZERO).neg()

ckamm (approved):

PR #446, opened 2023-02-09
fix log amount in token deposit

microwavedcola1:


programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -237,7 +237,7 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
             mango_account: self.account.key(),
             signer: self.token_authority.key(),
             token_index,
-            quantity: amount,
+            quantity: amount_i80f48.to_num::<u64>(),
```
ckamm, line 240: This is definitely more correct than the unadjusted `amount`, so approved. Let's discuss out-of-sync issues from fractional deposits separately.

ckamm (approved):

PR #447, opened 2023-02-09
Change liquidation end target to a new, third health type

ckamm:
Due to the safety features in v4, the init health can differ from maint health a lot more than it used to in v3. This is because of stable-price adjusted oracle prices used in init health, and the weight scaling based on total deposits and borrows used in init health.

The effect is that once an account becomes liquidatable, it could be liquidated a lot until it reaches init>=0.

The original idea of liquidating until init>=0 was just to provide some buffer, such that liquidated accounts wouldn't immediately become liquidatable again.

This patch decouples the buffer idea explicit from init health by creating a new LiquidationEnd health type. Liquidation proceeds until the LiquidationEnd health becomes positive.

### DIFF
--- a/programs/mango-v4/src/health/cache.rs
+++ b/programs/mango-v4/src/health/cache.rs
@@ -728,14 +728,17 @@ pub fn new_health_cache(
             oracle: oracle_price,
             stable: bank.stable_price(),
         };
+        // Use the liab price for computing weight scaling, because it's pessimistic and
+        // causes the most unfavorable scaling.
+        let liab_price = prices.liab(HealthType::Init);
         token_infos.push(TokenInfo {
             token_index: bank.token_index,
             maint_asset_weight: bank.maint_asset_weight,
             init_asset_weight: bank.init_asset_weight,
-            init_scaled_asset_weight: bank.scaled_init_asset_weight(prices.asset(HealthType::Init)),
+            init_scaled_asset_weight: bank.scaled_init_asset_weight(liab_price),
             maint_liab_weight: bank.maint_liab_weight,
             init_liab_weight: bank.init_liab_weight,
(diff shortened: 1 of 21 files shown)


programs/mango-v4/src/instructions/serum3_liq_force_cancel_orders.rs
```diff
@@ -241,7 +241,7 @@ pub fn serum3_liq_force_cancel_orders(
     //
     // Health check at the end
     //
-    let init_health = health_cache.health(HealthType::Init);
+    let init_health = health_cache.health(HealthType::LiquidationEnd);
```
ckamm, line 244: bad name

programs/mango-v4/src/health/cache.rs
```diff
@@ -717,9 +731,11 @@ pub fn new_health_cache(
         token_infos.push(TokenInfo {
             token_index: bank.token_index,
             maint_asset_weight: bank.maint_asset_weight,
-            init_asset_weight: bank.scaled_init_asset_weight(prices.asset(HealthType::Init)),
+            init_asset_weight: bank.init_asset_weight,
+            init_scaled_asset_weight: bank.scaled_init_asset_weight(prices.asset(HealthType::Init)),
```
microwavedcola1, line 735: ok I just noticed this, why is deposit_weight_scale_start_quote compared with deposits computed using potentially stable price? hmm shouldn't it use raw oracle price always?

same for scaled_init_liab_weight below

ckamm: Mmh. Could argue that using the oracle price leads to stricter results for asset weight scaling (because the stable price can only reduce the combined asset price). For liab weight the combined price leads to stricter results because it can only be >= oracle.

I'll make the change.

microwavedcola1 (approved):

PR #454, opened 2023-02-14
License: Mark instructions GPLv3 and everything else MIT

ckamm:
To do that, split up the Accounts objects and the instruction implementations.

GPL code is only used when the "enable-gpl" feature is enabled. That means compiling the program or running tests need explicit feature activation now.


programs/mango-v4/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused_variables)]
```
microwavedcola1, line 1: curious, why was this needed?

microwavedcola1 (approved):

programs/mango-v4/src/lib.rs
ckamm: When compiled without `enable-gpl`, none of the instruction function arguments will end up being used - unfortunately I didn't find a way to conditionally enable this flag.

PR #455, opened 2023-02-14
dont try to update mngo-perp-old since its in reduce only with no oi

microwavedcola1:


ckamm (approved):
While this is fine, maybe the generic rule "don't crank if it's reduce-only with zero open interest" also makes sense to me

PR #460, opened 2023-02-15
fix fillevent type in client

tylersssss:
fix fillevent type in client


tylersssss:
ui relies on this type so it would be helpful to get it merged quickly

PR #461, opened 2023-02-16
rust client: log transaction error along with logs

microwavedcola1:


client/src/client.rs
```diff
@@ -1554,7 +1554,10 @@ pub fn prettify_solana_client_error(
             RpcResponseErrorData::SendTransactionPreflightFailure(s) => {
                 if let Some(logs) = s.logs.as_ref() {
                     return MangoClientError::SendTransactionPreflightFailure {
-                        logs: logs.iter().join("; "),
+                        logs: logs.iter().join("; ")
+                            + &s.err
+                                .as_ref()
+                                .map_or_else(|| "".to_owned(), |e| e.to_string()),
```
ckamm, line 1560: This is too much of a hack for me: we wouldn't get the info we need when logs is None. 

Either we build a stringized representation of the error here that may contain log and error data if it exists - or maybe we store copy of the RpcSimulateTransactionResult our version of SendTransactionPreflightFailure and do nicer formatting later.

PR #464, opened 2023-02-17
buyback perp fees with mngo at a discount

microwavedcola1:


programs/mango-v4/tests/cases/test_perp.rs
```diff
@@ -456,10 +456,68 @@ async fn test_perp_fixed() -> Result<(), TransportError> {
...
+        (mngo_token_position_before - mngo_token_position_after).to_num::<f64>()
+            - 0.031999999668447
+            < 0.00000001
+    );
+
     Ok(())
 }
 
-#[tokio::test]
+// #[tokio::test]
```
microwavedcola1, line 520: todo undo

programs/mango-v4/src/instructions/account_settle_fees_accrued_with_mngo.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    let mut account = ctx.accounts.account.load_full_mut()?;
+    // account constraint #1
+    require!(
+        account.fixed.is_owner_or_delegate(ctx.accounts.owner.key()),
+        MangoError::SomeError
+    );
+    let mut dao_account = ctx.accounts.dao_account.load_full_mut()?;
+
+    // check positions exist, for nicer error messages
```
ckamm, line 35: I'd use enforce_token_position here. Maybe the USDC position on the dao_account or user account got emptied and implicitly closed.

The current code makes sense for settle_pnl because the USDC position is marked as "in_use" for the time that the user has open perp positions.

programs/mango-v4/src/instructions/account_settle_fees_accrued_with_mngo.rs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        account.token_position(mngo_bank.token_index)?;
+        account.token_position(settle_bank.token_index)?;
+        dao_account.token_position(mngo_bank.token_index)?;
+        dao_account.token_position(settle_bank.token_index)?;
+    }
+
+    let discount_rate = I80F48::from_num::<f32>(group.fees_mngo_discount_rate);
```
ckamm, line 43: shouldn't have to be explicit about f32 here

programs/mango-v4/src/state/mango_account.rs
```diff
@@ -88,7 +88,9 @@ pub struct MangoAccount {
 
     pub frozen_until: u64,
 
-    pub reserved: [u8; 232],
+    pub fees_accrued: u64,
```
ckamm, line 91: I don't like the name, since it's not tracking the overall fees_accrued at all. Instead it tracks only some fees. And it gets decreased when the MNGO->USDC conversion gets used.

programs/mango-v4/src/accounts_ix/account_settle_fees_accrued_with_mngo.rs
```diff
@@ -0,0 +1,47 @@
+use crate::error::*;
+use crate::state::*;
+use anchor_lang::prelude::*;
+
+#[derive(Accounts)]
+pub struct AccountSettleFeesAccruedWithMngo<'info> {
```
ckamm, line 6: maybe remove "Accrued" from the name?

programs/mango-v4/src/instructions/account_settle_fees_accrued_with_mngo.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+    let mngo_oracle_price = mngo_bank.oracle_price(
+        &AccountInfoRef::borrow(&ctx.accounts.mngo_oracle.as_ref())?,
+        None,
+    )?;
+    // mngo is exchanged at a discount
+    let mut max_settle_mngo = cm!(max_settle * discount_rate / mngo_oracle_price);
+    // settlement is restricted to accounts token position
+    max_settle_mngo = max_settle_mngo.min(account_mngo_native);
+    max_settle = cm!(max_settle_mngo * mngo_oracle_price / discount_rate);
```
ckamm, line 66: If we stored the inverse of discount rate, we could skip one division:
```
let mngo_settle_price = cm!(mngo_oracle_price * price_bonus);`
```
and then divide and multiply with mngo_settle_price instead

programs/mango-v4/src/instructions/account_settle_fees_accrued_with_mngo.rs
```diff
@@ -0,0 +1,98 @@
...
+    // move settlement tokens from dao to user
+    let account_settle_token_position = account.token_position_mut(settle_bank.token_index)?.0;
+    let dao_settle_token_position = dao_account.token_position_mut(settle_bank.token_index)?.0;
+    settle_bank.withdraw_without_fee(
+        dao_settle_token_position,
+        max_settle,
+        now_ts,
+        mngo_oracle_price,
+    )?;
+    settle_bank.deposit(account_settle_token_position, max_settle, now_ts)?;
```
ckamm, line 87: these four balance changes might implicitly close any of these four token positions, you need to track the return values and possibly close deactivate_token_position_... :(

programs/mango-v4/src/instructions/account_settle_fees_accrued_with_mngo.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    account.fixed.fees_accrued =
+        cm!(account.fixed.fees_accrued - max_settle.round().to_num::<u64>());
```
ckamm, line 90: Rounding could lead to an exploit. Ceil instead, but also use saturating_sub so there's no potential panic here if there's a path where rounding errors make max_settle slightly bigger than fees_accrued.

ckamm (commented):
initial comments, need to re-read

programs/mango-v4/src/state/orderbook/book.rs
```diff
@@ -363,6 +364,8 @@ fn apply_fees(
     require_gte!(taker_fees, 0);
 
     // The maker fees apply to the maker's account only when the fill event is consumed.
+    account.fixed.discount_settleable_fees_accrued += taker_fees.round().to_num::<u64>();
```
ckamm, line 367: let's also add the maker fees if they are not negative, in perp_consume_events

programs/mango-v4/tests/cases/test_perp.rs
```diff
@@ -23,7 +23,7 @@ async fn test_perp_fixed() -> Result<(), TransportError> {
     .create(solana)
     .await;
 
-    let deposit_amount = 1000;
+    let deposit_amount = 1000_000;
```
ckamm, line 26: undo?

programs/mango-v4/tests/program_test/mango_client.rs
```diff
@@ -1503,6 +1503,49 @@ impl ClientInstruction for GroupCreateInstruction {
     }
 }
 
+pub struct GroupEditInstruction {
```
ckamm, line 1506: rename to GroupEditFeeParameters or something?

programs/mango-v4/tests/program_test/mango_client.rs
```diff
@@ -1503,6 +1503,49 @@ impl ClientInstruction for GroupCreateInstruction {
...
+        _account_loader: impl ClientAccountLoader + 'async_trait,
+    ) -> (Self::Accounts, instruction::Instruction) {
+        let program_id = mango_v4::id();
+        let instruction = Self::Instruction {
+            admin_opt: None,
+            fast_listing_admin_opt: None,
+            security_admin_opt: None,
+            testing_opt: None,
+            version_opt: None,
+            deposit_limit_quote_opt: None,
```
ckamm, line 1528: recommend making a function that returns a all-None edit instruction, like for perp_edit

programs/mango-v4/tests/program_test/mango_client.rs
```diff
@@ -1765,6 +1808,49 @@ impl ClientInstruction for AccountCloseInstruction {
     }
 }
 
+pub struct AccountSettleFeesWithMngo {
+    pub group: Pubkey,
+    pub owner: TestKeypair,
+    pub account: Pubkey,
+    pub dao_account: Pubkey,
```
ckamm, line 1815: no need to pass the group or dao_account explicitly, can get the group key from the account, load it and fetch the dao_account

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -32,6 +32,7 @@ export class MangoAccount {
       perpSpotTransfers: BN;
       healthRegionBeginInitHealth: BN;
       frozenUntil: BN;
+      feesAccrued: BN;
```
ckamm, line 35: this still calls it `feesAccrued`

programs/mango-v4/src/instructions/account_settle_fees_accrued_with_mngo.rs
microwavedcola1: would keeping these positions open be an issue?

ckamm: Not directly. The main two reasons are consistency - the other instructions close token positions that were dusted, not doing so is kind of an invalid state. And secondly if the user's MNGO token position stays open, then the user needs a workaround to close it when he wants to reuse the token index (something like depositing 1 native and withdrawing it again)

But yeah, I'm pretty annoyed about having to do this in all instructions that deal with token positions. Possibly it'd be better to skip it and auto-close zeroed token positions in enforce_token_position? That also has some implications for which banks/oracles need to be send for health checks though.

programs/mango-v4/tests/cases/test_fees_settle_with_mngo.rs
```diff
@@ -0,0 +1,187 @@
...
+    assert_eq!(before_fees_accrued - after_fees_accrued, 20);
+
+    // token[1] swapped at discount for token[0]
+    assert!(
+        (settle_token_position_after - settle_token_position_before) - I80F48::from_num(20)
+            < I80F48::from_num(0.000001)
+    );
+    assert!(
+        (mngo_token_position_before - mngo_token_position_after) - I80F48::from_num(16.666666)
+            < I80F48::from_num(0.000001)
```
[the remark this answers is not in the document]
ckamm: ok

programs/mango-v4/src/instructions/account_settle_fees_accrued_with_mngo.rs
microwavedcola1: I have now deactivated the token positions if not in_use after withdraw, please check latest commit, afaics positions where deposit is happening don't need to be checked

I have also added an additional check which dao fees token position >= max_buyback

microwavedcola1: I have also added deactivations for deposits, to just be consistent

programs/mango-v4/src/accounts_ix/account_buyback_fees_with_mngo.rs
```diff
@@ -0,0 +1,52 @@
...
+        // owner is checked at #1
+    )]
+    pub account: AccountLoader<'info, MangoAccountFixed>,
+    pub owner: Signer<'info>,
+
+    #[account(
+        mut,
+        has_one = group,
+        constraint = account.load()?.is_operational() @ MangoError::AccountIsFrozen,
+        constraint = group.load()?.fees_swap_mango_account == dao_account.key()
```
ckamm, line 26: minor: could use an `address =` constraint instead

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -0,0 +1,151 @@
...
+    );
+
+    let mut dao_account = ctx.accounts.dao_account.load_full_mut()?;
+
+    let group = ctx.accounts.group.load()?;
+
+    let mut mngo_bank = ctx.accounts.mngo_bank.load_mut()?;
+    let mut fees_bank = ctx.accounts.fees_bank.load_mut()?;
+
+    let bonus_rate = I80F48::from_num(group.fees_mngo_bonus_factor);
```
ckamm, line 36: -> bonus_factor

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    // compute max mngo to swap for fees
+    let mngo_oracle_price = mngo_bank.oracle_price(
+        &AccountInfoRef::borrow(&ctx.accounts.mngo_oracle.as_ref())?,
+        Some(Clock::get()?.slot),
+    )?;
+    let mngo_buyback_price = cm!(mngo_oracle_price * bonus_rate);
+    // mngo is exchanged at a discount
+    let mut max_buyback_mngo = cm!(max_buyback / mngo_buyback_price);
+    // buyback is restricted to accounts token position
```
ckamm, line 70: account's

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    // ensure dao mango account has no liabilities after we do the token swap
+    for ele in dao_account.all_token_positions() {
+        require!(!ele.indexed_position.is_negative(), MangoError::SomeError);
+    }
+    require_eq!(
+        dao_account.active_perp_positions().count(),
+        0,
+        MangoError::SomeError
+    );
```
ckamm, line 148: wouldn't hurt to also check "no serum open orders"

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    let mut mngo_bank = ctx.accounts.mngo_bank.load_mut()?;
+    let mut fees_bank = ctx.accounts.fees_bank.load_mut()?;
+
+    let bonus_rate = I80F48::from_num(group.fees_mngo_bonus_factor);
+    let now_ts = Clock::get()?.unix_timestamp.try_into().unwrap();
+
+    // quick return if nothing to buyback
+    let mut max_buyback =
+        I80F48::from_num::<u64>(max_buyback.min(account.fixed.discount_buyback_fees_accrued));
```
ckamm, line 41: Imo this should also min() with the available native USDC balance on the dao account. But technically we shouldn't do any mutations (enforce_token_pos) before the early-return here. Maybe we just don't care about token position open/closed state on the dao_account? (do care for the user account though)

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -0,0 +1,151 @@
...
+        mngo_oracle_price,
+    )?;
+    if !in_use {
+        account.deactivate_token_position_and_log(
+            account_mngo_raw_token_index,
+            ctx.accounts.account.key(),
+        );
+    }
+    let in_use = mngo_bank.deposit(dao_mngo_token_position, max_buyback_mngo, now_ts)?;
+    if !in_use {
```
ckamm, line 90: this `in_use` check we could skip by requiring that the dao_account's mngo token position is >= 0 beforehand

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    // move fees from dao to user
+    let (account_fees_token_position, account_fees_raw_token_index, _) =
+        account.ensure_token_position(fees_bank.token_index)?;
+    let (dao_fees_token_position, dao_fees_raw_token_index, _) =
+        dao_account.ensure_token_position(fees_bank.token_index)?;
+    let dao_fees_native = dao_fees_token_position.native(&fees_bank);
+    if dao_fees_native.is_negative() || dao_fees_native < max_buyback {
+        msg!(
+            "dao fees token position ({} native fees) is lesser than max buyback {}, nothing will be bought back",
```
ckamm, line 105: see comment above, I think the buyback should always be limited to dao_fees_native

programs/mango-v4/src/instructions/group_edit.rs
```diff
@@ -12,6 +13,10 @@ pub fn group_edit(
     testing_opt: Option<u8>,
     version_opt: Option<u8>,
     deposit_limit_quote_opt: Option<u64>,
+    fees_pay_with_mngo_opt: Option<bool>,
+    fees_mngo_bonus_factor_opt: Option<f32>,
+    fees_swap_mango_account_opt: Option<Pubkey>,
```
ckamm, line 18: let's have a consistent name for this, I've seen:
- buyback_fees
- discount_buyback_fees
- fees_pay_with_mngo
- fees_swap

I'd go for `buyback_fees` I think

programs/mango-v4/src/instructions/group_edit.rs
```diff
@@ -12,6 +13,10 @@ pub fn group_edit(
...
+    fees_mngo_token_index_opt: Option<TokenIndex>,
```
ckamm, line 19: this one shouldn't be specific to buyback_fees, just `mngo_token_index` seems ok

programs/mango-v4/src/state/mango_account.rs
```diff
@@ -899,6 +906,10 @@ impl<
         let (base_change, quote_change) = fill.base_quote_change(side);
         let quote = cm!(I80F48::from(perp_market.quote_lot_size) * I80F48::from(quote_change));
         let fees = cm!(quote.abs() * I80F48::from_num(fill.maker_fee));
+        if fees.is_positive() {
+            self.fixed_mut().discount_buyback_fees_accrued += fees.round().to_num::<u64>();
```
ckamm, line 910: personally, I'd floor() here but it's not a big deal

programs/mango-v4/src/state/mango_account.rs
```diff
@@ -899,6 +906,10 @@ impl<
...
+        }
+        let pa = self.perp_position_mut(perp_market_index)?;
```
ckamm, line 912: looks like `pa` is unused above, move the whole thing (including settle_funding) here?

programs/mango-v4/src/state/orderbook/book.rs
```diff
@@ -363,6 +364,8 @@ fn apply_fees(
     require_gte!(taker_fees, 0);
 
     // The maker fees apply to the maker's account only when the fill event is consumed.
+    account.fixed.discount_buyback_fees_accrued += taker_fees.round().to_num::<u64>();
```
ckamm, line 367: maybe: floor()?

programs/mango-v4/src/state/orderbook/book.rs
```diff
@@ -375,8 +378,11 @@ fn apply_fees(
 }
 
 /// Applies a fixed penalty fee to the account, and update the market's fees_accrued
-fn apply_penalty(market: &mut PerpMarket, perp_position: &mut PerpPosition) -> Result<()> {
+fn apply_penalty(market: &mut PerpMarket, account: &mut MangoAccountRefMut) -> Result<()> {
     let fee_penalty = I80F48::from_num(market.fee_penalty);
+    account.fixed.discount_buyback_fees_accrued += fee_penalty.round().to_num::<u64>();
```
ckamm, line 383: floor?

programs/mango-v4/src/instructions/group_edit.rs
microwavedcola1: all done, please recheck though

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    let bonus_factor = I80F48::from_num(group.buyback_fees_mngo_bonus_factor);
+    let now_ts = Clock::get()?.unix_timestamp.try_into().unwrap();
+
+    // quick return if nothing to buyback
+    let mut max_buyback = {
+        let dao_fees_token_position = dao_account.ensure_token_position(fees_bank.token_index)?.0;
+        let dao_fees_native = dao_fees_token_position.native(&fees_bank);
+        I80F48::from_num::<u64>(max_buyback.min(account.fixed.buyback_fees_accrued))
+            .min(dao_fees_native)
```
ckamm, line 44: I'm worried about what happens here if dao_fees_native is negative. We need to either require it's not, or change the check below to max_buyback <= 0 (the latter would be more defensive anyway)

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -0,0 +1,158 @@
...
+            account.fixed.buyback_fees_accrued
+        );
+        return Ok(());
+    }
+
+    // if mngo token position has borrows, skip buyback
+    let (account_mngo_token_position, account_mngo_raw_token_index, _) =
+        account.ensure_token_position(mngo_bank.token_index)?;
+    let account_mngo_native = account_mngo_token_position.native(&mngo_bank);
+    if account_mngo_native.is_negative() {
```
ckamm, line 58: could we early out on <=0? That way we don't have the special case where max_buyback = max_buyback_mngo = 0 below. Erroring about "has no deposits"...

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+    mngo_bank.deposit(dao_mngo_token_position, max_buyback_mngo, now_ts)?;
+
+    // move fees from dao to user
+    let (account_fees_token_position, account_fees_raw_token_index, _) =
+        account.ensure_token_position(fees_bank.token_index)?;
+    let (dao_fees_token_position, dao_fees_raw_token_index, _) =
+        dao_account.ensure_token_position(fees_bank.token_index)?;
+    let dao_fees_native = dao_fees_token_position.native(&fees_bank);
+    if dao_fees_native.is_negative() || dao_fees_native < max_buyback {
```
ckamm, line 105: could be converted to an assert!() because the above math + checks guarantee that this can never be true

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -0,0 +1,158 @@
...
+        max_buyback,
+        max_buyback_mngo
+    );
+
+    // ensure dao mango account has no liabilities after we do the token swap
+    for ele in dao_account.all_token_positions() {
+        require!(!ele.indexed_position.is_negative(), MangoError::SomeError);
+    }
+    require_eq!(
+        dao_account.active_perp_positions().count(),
```
ckamm, line 147: typo, one of these should be serum positions

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -55,9 +54,9 @@ pub fn account_buyback_fees_with_mngo(
     let (account_mngo_token_position, account_mngo_raw_token_index, _) =
         account.ensure_token_position(mngo_bank.token_index)?;
     let account_mngo_native = account_mngo_token_position.native(&mngo_bank);
-    if account_mngo_native.is_negative() {
+    if account_mngo_native <= I80F48::ZERO {
         msg!(
-            "account mngo token position ({} native mngo) has borrows, nothing will be bought back",
+            "account mngo token position ({} native mngo) is <= 0, nothing will be bought back",
```
ckamm, line 59: Gah, I realized we have a problem here now. If account_mngo_native == 0, then it could be that this is a newly created mango token position on `account`. When we `return Ok(())` that gets committed. So now the user has a active but zero position. 

Maybe we do something like
```
let account_mngo_native = account.token_position(mngo_bank.token_index)
    .map(|tp| tp.native(&mngo_bank))
    .unwrap_or(I80F48::ZERO);
```
so we have no pending changes when the early-return happens?

ckamm (approved):

PR #465, opened 2023-02-17
Add a "settle-bot" binary

ckamm:
The bot will call settle on all accounts with unsettled pnl where it can receive either the flat or low-health fee.


settle-bot/Dockerfile.liquidator
```diff
@@ -0,0 +1,7 @@
+# Dockerfile for keeper service in Heroku
```
microwavedcola1, line 1: probably also want to copy paste stuff to push settle bot to gcr, see ci-docker-publish.yml, if it doesnt work out of the box, we could ask pan of course

settle-bot/.env.example
```diff
@@ -0,0 +1,4 @@
+RPC_URL=https://mango.devnet.rpcpool.com
```
microwavedcola1, line 1: these need to be updated or?

settle-bot/src/settle.rs
```diff
@@ -0,0 +1,315 @@
...
+                {
+                    // no more interesting pairs that would produce fees
+                    break;
+                }
+
+                // add to batch and maybe send
+                batch_processor
+                    .add_and_maybe_send(account_a, account_b)
+                    .await?;
+                negative_settleable.change_priority(&account_b, settleable_b - settleable);
```
microwavedcola1, line 211: hmm do we keep on reusing account_a , settleable_a, without shaving off settleable_a after we settle..similar to how we do for account_b?

ckamm: I'll add a note about that. We loop through the positive-pnl accounts exactly once, so modifying these values doesn't matter.

microwavedcola1: ah, but you could have a mm, who has a large position and be +ve pnl, means the distribution won't be smooth right

ckamm: Yeah, if there's an account with lots of +pnl and all accounts with -pnl are small then that could take multiple passes to get settled right now. I'll add that to the note.

ckamm: actually did a nice solution now

microwavedcola1 (approved):

PR #472, opened 2023-02-22
security admin can move openbook market to reduce only

microwavedcola1:


programs/mango-v4/src/instructions/serum3_edit_market.rs
```diff
@@ -15,5 +18,19 @@ pub fn serum3_edit_market(
         );
         serum3_market.reduce_only = u8::from(reduce_only);
```
ckamm, line 19: add
```
        // security admin can only enable reduce_only
        if !reduce_only {
            require_group_admin = true;
        }
 ```

ckamm (approved):

PR #480, opened 2023-02-27
ts client for max buyback of fees with mngo

microwavedcola1:


ts/client/src/accounts/group.ts
```diff
@@ -466,6 +466,18 @@ export class Group {
...
+  public getFirstBankForMngo(): Bank {
+    return this.getFirstBankByTokenIndex(this.mngoTokenIndex);
+  }
+
+  public getFirstBankForFees(): Bank {
+    return this.getFirstBankByTokenIndex(0);
+  }
+
+  public getFirstBankForSettlement(): Bank {
+    return this.getFirstBankByTokenIndex(0);
```
ckamm, line 478: Maybe be more specific and say "PerpSettlement"? Not a strong opinion though.

"ForFees" I don't get and seems strictly too generic. Fees can apply on any bank after all.

ckamm (approved):

PR #484, opened 2023-02-28
Serum settle funds V2: fees can go to users

ckamm:


programs/mango-v4/src/instructions/serum3_place_order.rs
```diff
@@ -424,22 +424,26 @@ pub fn apply_settle_changes(
     after_quote_vault: u64,
     after_oo: &OpenOrdersSlim,
     health_cache: Option<&mut HealthCache>,
+    fees_to_dao: bool,
```
microwavedcola1, line 427: is apply_settle_changes not called in this file at all? why is it in this file then?

microwavedcola1 (approved):

programs/mango-v4/src/instructions/serum3_place_order.rs
ckamm: Only because apply_settle_changes() works via apply_vault_difference() and the latter is only needed in place_order. If you don't feel too strongly about this, I'd leave it as-is.

ts/client/src/client.ts
```diff
@@ -1696,6 +1698,81 @@ export class MangoClient {
...
+            serum3Market.baseTokenIndex,
+          ).oracle,
+        },
+      })
+      .instruction();
+
+    return ix;
+  }
+
+  public async serum3SettleFundsV2(
```
ckamm, line 1762: I'm wondering how to migrate here. Proposal and thoughts:
- Switching to the new instruction everywhere would cause the client code to depend on the new program deployment for a while, I'd rather avoid that.
- Change serum3SettleFunds() and serum3SettleFundsIx() to use V2 instructions when openbookFeesToDao==false. That way user code works correctly without changes.
- Possibly introduce serum3SettleFundsV1Ix() that asserts if openbookFeesToDao is false?
- It'd be cool if the client could determine the deployed program verison somehow and use that to decide what behaviors are supported. Is it easy to fetch the "version" field from the on-chain program idl?

ckamm: Additionally, we may want to prefer the v2 instruction globally at some point, if we can give the user additional fee-buyback benefits now that the oracles are passed in.

microwavedcola1: ok so I will

* remove the serum3SettleFundsV2Ix and serum3SettleFundsV2
* make serum3SettleFundsIx and serum3SettleFunds use v1 ix if flag is true, use v2 if flag is false
* yes clients can determine, but would need to fetch on chain idl, maybe can do this async, if we see idl with version where v2 ix is present, then always use v2 ix, wdyt?

PR #488, opened 2023-03-02
support name edit for token and program

microwavedcola1:


programs/mango-v4/src/instructions/perp_edit_market.rs
```diff
@@ -322,6 +324,12 @@ pub fn perp_edit_market(
         require_group_admin = true;
     }
 
+    if let Some(name) = name_opt.clone() {
```
ckamm, line 327: why's the clone necessary? should definitely be able to as_ref() instead

ckamm (approved):

PR #491, opened 2023-03-03
Add oracleProvider to Bank and PerpMarket

riordanp:
Adds a string representing the oracle provider for use in the UI


ts/client/src/accounts/group.ts
```diff
@@ -395,11 +402,13 @@ export class Group {
       price = new I80F48(stubOracle.price.val);
       uiPrice = this.toUiPrice(price, baseDecimals);
       lastUpdatedSlot = stubOracle.lastUpdated.val;
+      provider = 'stub';
```
microwavedcola1, line 405: make enum?

microwavedcola1 (approved):

PR #503, opened 2023-03-15
token_liq_bankruptcy: Use oracle for valuing insurance fund tokens

ckamm:
Previously a token from the insurance fund was valued at 1 USD. Now it uses the oracle associated with it (USDC oracle).

### DIFF
--- a/programs/mango-v4/src/state/group.rs
+++ b/programs/mango-v4/src/state/group.rs
@@ -4,7 +4,24 @@ use std::mem::size_of;
 
 // TODO: Assuming we allow up to 65536 different tokens
 pub type TokenIndex = u16;
+
+/// This token index is supposed to be the token that oracles quote in.
+///
+/// In practice this is set to the USDC token index, and that is wrong: actually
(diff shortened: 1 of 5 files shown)


programs/mango-v4/src/state/group.rs
```diff
@@ -4,8 +4,25 @@ use std::mem::size_of;
 
 // TODO: Assuming we allow up to 65536 different tokens
 pub type TokenIndex = u16;
+
+/// This token index is supposed to be the token that oracles quote in.
+///
+/// In practice this is set to the USDC token index, and that is wrong: actually
```
microwavedcola1, line 10: so this comment will go away, when we do the big change, right? when the quoting/tick comes in usdc

microwavedcola1 (approved):

programs/mango-v4/src/state/group.rs
ckamm: The whole QUOTE_TOKEN_INDEX constant needs to go away. There's no registered token that matches the oracle quote currency.

PR #504, opened 2023-03-15
Fee buyback: Respect USDC oracle

ckamm:


programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -34,21 +34,36 @@ pub fn account_buyback_fees_with_mngo(
...
     let bonus_factor = I80F48::from_num(group.buyback_fees_mngo_bonus_factor);
 
-    let now_ts = Clock::get()?.unix_timestamp.try_into().unwrap();
     account
         .fixed
         .expire_buyback_fees(now_ts, group.buyback_fees_expiry_interval);
 
     // quick return if nothing to buyback
-    let mut max_buyback = {
+    let mut max_buyback_usd_native = {
```
microwavedcola1, line 60: I thought we agreed that if no suffix then its native

microwavedcola1 (approved):

programs/mango-v4/src/lib.rs
```diff
@@ -285,10 +285,10 @@ pub mod mango_v4 {
 
     pub fn account_buyback_fees_with_mngo(
         ctx: Context<AccountBuybackFeesWithMngo>,
-        max_buyback: u64,
+        max_buyback_usd_native: u64,
```
microwavedcola1, line 288: nit: could you also match the arg in ts/client

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
ckamm: right, I just did it because the other fields had a _native postfix in that function. Maybe I'll get rid of all of them.

microwavedcola1 (approved):

PR #506, opened 2023-03-16
wip: extract perp order id from tx

microwavedcola1:


ts/client/src/client.ts
```diff
@@ -2083,6 +2086,52 @@ export class MangoClient {
...
+
+      const returnPrefix = `Program return: ${client.programId} `;
+      const returnLogs = tx?.meta?.logMessages?.filter((l) =>
+        l.startsWith(returnPrefix),
+      );
+      if (!returnLogs) {
+        throw new Error('Expected return log');
+      }
+
+      return returnLogs.map((returnLog) => {
```
ckamm, line 2117: This should only be done for instructions that actually call the ixName instruction. Otherwise this would fail if another v4 instruction was called and also returned data, I think?

PR #508, opened 2023-03-20
HealthRegion: Whitelist allowed instruction types

ckamm:
This fixes a security issue where bankruptcy related instructions could be called inside a health region. Now health regions are limited to compute optimization like when placing multiple orders in one transaction.

This limitation also makes it impossible to abuse health regions for flash loans. Use the FlashLoan instructions for that purpose.


programs/mango-v4/src/instructions/health_region.rs
```diff
@@ -10,11 +10,33 @@ use fixed::types::I80F48;
...
+        crate::instruction::PerpCancelAllOrdersBySide::discriminator(),
+        crate::instruction::PerpCancelOrder::discriminator(),
+        crate::instruction::PerpCancelOrderByClientOrderId::discriminator(),
+        crate::instruction::PerpPlaceOrder::discriminator(),
+        crate::instruction::PerpPlaceOrderPegged::discriminator(),
+        crate::instruction::Serum3CancelAllOrders::discriminator(),
+        crate::instruction::Serum3CancelOrder::discriminator(),
+        crate::instruction::Serum3PlaceOrder::discriminator(),
+        crate::instruction::Serum3SettleFunds::discriminator(),
+        crate::instruction::Serum3SettleFundsV2::discriminator(),
```
microwavedcola1, line 25: Similar to above, how about create oo and close oo orders ixs?

microwavedcola1 (approved):

PR #518, opened 2023-03-24
force close tokens program part

microwavedcola1:


programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,200 @@
...
+    require_msg_typed!(
+        !liqor.fixed.being_liquidated(),
+        MangoError::BeingLiquidated,
+        "liqor account"
+    );
+
+    let mut liqee = ctx.accounts.liqee.load_full_mut()?;
+    let mut liqee_health_cache = new_health_cache(&liqee.borrow(), &account_retriever)
+        .context("create liqee health cache")?;
+    let liqee_liq_end_health = liqee_health_cache.health(HealthType::Maint);
```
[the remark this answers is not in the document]
microwavedcola1: no this is a mistake only here afaik, will switch to HealthType::LiquidationEnd

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,200 @@
...
+use crate::error::*;
+use crate::health::*;
+use crate::logs::TokenBalanceLog;
+use crate::state::*;
+
+// TODO: alternative ix name suggestions?
+pub fn token_force_close_borrows_with_token(
+    ctx: Context<TokenForceCloseBorrowsWithToken>,
+    asset_token_index: TokenIndex,
+    // token's mode is checked at #2
```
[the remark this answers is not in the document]
microwavedcola1: I imagine force close as also a mode, do you mean I should 
* also check reduce only is true or
* only check reduce only is true or
* keep check to force close is true && enforce that force close can only be set to true on token edit if reduce only is already true
?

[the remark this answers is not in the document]
ckamm: I think checking for force_close is sufficient in this instruction

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,200 @@
...
+        let (liqee_liab_position, liqee_liab_raw_index) =
+            liqee.token_position_and_raw_index(liab_token_index)?;
+        let liqee_liab_native = liqee_liab_position.native(liab_bank);
+        require!(liqee_liab_native.is_negative(), MangoError::SomeError);
+
+        // The amount of liab native tokens we will transfer
+        let liab_transfer = min(-liqee_liab_native, max_liab_transfer);
+
+        // The amount of asset native tokens we will give up for them
+        let fee_factor = I80F48::ONE + asset_bank.liquidation_fee + liab_bank.liquidation_fee;
```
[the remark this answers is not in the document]
ckamm: This probably needs to propagate to other liquidation instructions.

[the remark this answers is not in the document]
microwavedcola1: done in 
image

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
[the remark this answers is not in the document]
ckamm: I'd still nitpick that "token's force_close flag is checked at # 2" would confuse me less

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,207 @@
...
+    ctx: Context<TokenForceCloseBorrowsWithToken>,
+    asset_token_index: TokenIndex,
+    // token's mode is checked at #2
+    liab_token_index: TokenIndex,
+    max_liab_transfer: I80F48,
+) -> Result<()> {
+    let group_pk = &ctx.accounts.group.key();
+
+    require_neq!(asset_token_index, liab_token_index, MangoError::SomeError);
+    // TODO: should we enforce that asset token index is always USDC?
```
ckamm, line 22: Hmm, arguments could be made in both directions:
- We don't want force-closers to create random new borrows on users (use asset_token = BONK)
- We don't want to make USDC's role special

We could arbitrarily allow any token with liab weight <= 1.1? But that's also surprising. Hmm

ckamm: Any token with `asset_token_init_liab_weight <= liab_token_init_liab_weight + liab_token_liq_fee`? That way the liqee's health would be guaranteed to not decrease?

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    //
+    // Transfer liab_token from liqor to liqee to close the borrows.
+    // Transfer corresponding amount of asset_token from liqee to liqor.
+    //
+    let now_ts = Clock::get()?.unix_timestamp.try_into().unwrap();
+    {
+        let account_retriever: &mut ScanningAccountRetriever = &mut account_retriever;
+        let liab_token_index = liab_token_index;
+        let asset_token_index = asset_token_index;
```
ckamm, line 55: these three lines shouldn't do anything

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,207 @@
...
+        let liqor: &mut MangoAccountRefMut = &mut liqor.borrow_mut();
+        let liqor_key = ctx.accounts.liqor.key();
+        let liqee: &mut MangoAccountRefMut = &mut liqee.borrow_mut();
+        let liqee_key = ctx.accounts.liqee.key();
+
+        let (asset_bank, asset_oracle_price, opt_liab_bank_and_price) =
+            account_retriever.banks_mut_and_oracles(asset_token_index, liab_token_index)?;
+        let (liab_bank, liab_oracle_price) = opt_liab_bank_and_price.unwrap();
+        // account constraint #2
+        require!(liab_bank.is_force_close(), MangoError::SomeError);
```
ckamm, line 65: give this an error code or at least a require_msg?

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        let (liqee_asset_position, liqee_asset_raw_index) =
+            liqee.token_position_and_raw_index(asset_token_index)?;
+        let liqee_asset_native = liqee_asset_position.native(asset_bank);
+        // TODO: should we enforce that asset position is positive? or just incur borrows if needed, I think borrowing is fine
```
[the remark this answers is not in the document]
ckamm: remove comment

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        let (liqee_liab_position, liqee_liab_raw_index) =
+            liqee.token_position_and_raw_index(liab_token_index)?;
+        let liqee_liab_native = liqee_liab_position.native(liab_bank);
+        require!(liqee_liab_native.is_negative(), MangoError::SomeError);
+
+        // The amount of liab native tokens we will transfer
+        let liab_transfer = min(-liqee_liab_native, max_liab_transfer);
```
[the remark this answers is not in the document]
ckamm: I think liqor_liab_native should be min'ed into this number, the require_gte further down is too restrictive.

then also add .max(0) for safety

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        // The amount of asset native tokens we will give up for them
+        let fee_factor = I80F48::ONE + liab_bank.liquidation_fee;
+        let liab_oracle_price_adjusted = liab_oracle_price * fee_factor;
+        let asset_transfer = liab_transfer * liab_oracle_price_adjusted / asset_oracle_price;
+
+        // Apply the balance changes to the liqor and liqee accounts
+        let liqee_liab_position = liqee.token_position_mut_by_raw_index(liqee_liab_raw_index);
```
ckamm, line 86: why retrieve it again? we still have it in-scope! could we get the mut version from the start?

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        let (liqor_liab_position, liqor_liab_raw_index, _) =
+            liqor.ensure_token_position(liab_token_index)?;
+        let liqor_liab_native = liqor_liab_position.native(liab_bank);
+        // Require that liqor obtain deposits before he tries to liquidate liqee's borrows, to prevent
+        // moving other liquidator from earning further liquidation fee from these borrows by trying to liquidate the current liqor
+        require!(liqor_liab_native.is_positive(), MangoError::SomeError);
+        require_gte!(
+            liqor_liab_native.abs(),
+            liqee_liab_native.abs(),
```
ckamm, line 99: I don't like this because the caller may have specified a smaller max_settle -- better to fold this into the liab_transfer computation above

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,207 @@
...
+            deposit_index: liab_bank.deposit_index.to_bits(),
+            borrow_index: liab_bank.borrow_index.to_bits(),
+        });
+
+        // TODO: hmm should we just check if init_health was improved OR
+        // should we use this opportunity to get the liqee out of liquidated state?
+        let liqee_liq_end_health = liqee_health_cache.health(HealthType::LiquidationEnd);
+        liqee
+            .fixed
+            .maybe_recover_from_being_liquidated(liqee_liq_end_health);
```
ckamm, line 181: recovering from being liquidated makes sense to me

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,207 @@
...
+    require_msg_typed!(
+        !liqor.fixed.being_liquidated(),
+        MangoError::BeingLiquidated,
+        "liqor account"
+    );
+
+    let mut liqee = ctx.accounts.liqee.load_full_mut()?;
+    let mut liqee_health_cache = new_health_cache(&liqee.borrow(), &account_retriever)
+        .context("create liqee health cache")?;
+    let liqee_liq_end_health = liqee_health_cache.health(HealthType::LiquidationEnd);
```
ckamm, line 45: unclear why we compute this number here; if the whole liqee_health_cache is created further down, we don't need to bother adjusting it

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,207 @@
...
+        }
+        if !liqor_liab_active {
+            liqor.deactivate_token_position_and_log(liqor_liab_raw_index, liqor_key)
+        }
+    };
+
+    // Check liqor's health
+    let liqor_health = compute_health(&liqor.borrow(), HealthType::Init, &account_retriever)
+        .context("compute liqor health")?;
+    require!(liqor_health >= 0, MangoError::HealthMustBePositive);
```
ckamm, line 201: technically this should always improve liqor health, since we decrease the zero-asset-weight liab token and gain some asset token. but this check can stay for safety -- maybe leave a comment?

programs/mango-v4/src/state/bank.rs
```diff
@@ -127,10 +127,16 @@ pub struct Bank {
     /// See scaled_init_asset_weight().
     pub deposit_weight_scale_start_quote: f64,
 
+    // We have 3 modes
+    // 0 - Off,
+    // 1 - NoDepositsNoBorrows - standard
+    // 2 - NoBorrows - borrows can only be reduced, but deposits have no restriction, special case for
```
ckamm, line 133: minor: call this ReduceDepositsReduceBorrows and ReduceBorrows? You can still deposit to repay etc...

programs/mango-v4/tests/cases/test_force_close.rs
```diff
@@ -0,0 +1,275 @@
...
+        mints: mints.to_vec(),
+        ..GroupWithTokensConfig::default()
+    }
+    .create(solana)
+    .await;
+    let collateral_token = &tokens[0];
+    let borrow_token = &tokens[1];
+
+    // deposit some funds, to the vaults aren't empty
+    let vault_account = send_tx(
```
ckamm, line 31: use create_funded_account() for the vault, liqor and liqee; cuts down on boilerplate a good bit

programs/mango-v4/tests/program_test/mango_client.rs
```diff
@@ -1283,7 +1284,64 @@ impl ClientInstruction for TokenMakeReduceOnly {
...
+            }));
+        (accounts, instruction)
+    }
+
+    fn signers(&self) -> Vec<TestKeypair> {
+        vec![self.admin]
+    }
+}
+
+pub struct TokenMakeForceClose {
```
ckamm, line 1314: personally, I'd have joined this with TokenMakeReduceOnly since it's only used rarely and the two parameters are related...

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
microwavedcola1: makes sense, I'd use scaled weights or?

programs/mango-v4/src/instructions/token_edit.rs
```diff
@@ -277,8 +277,12 @@ pub fn token_edit(
                 bank.reduce_only,
                 reduce_only
             );
+
             // security admin can only make it stricter
-            if reduce_only < bank.reduce_only {
+            // anything that makes it less strict, should require admin
+            if (bank.reduce_only == 1 && reduce_only == 2)
+                || (bank.reduce_only == 2 && reduce_only == 0)
```
ckamm, line 284: this would mean the security admin could change to 0 if it's currently at 1.

Suggestion: `if reduce_only == 0 || (reduce_only == 2 && bank.reduce_only == 1)`
(require admin always for disabling; require admin for weakening from 1 to 2)

programs/mango-v4/src/instructions/flash_loan.rs
```diff
@@ -343,10 +343,15 @@ pub fn flash_loan_end<'key, 'accounts, 'remaining, 'info>(
 
         let change_amount = change.amount - loan_origination_fee;
         let native_after_change = native + change_amount;
-        if bank.is_reduce_only() {
+        if bank.are_deposits_reduce_only() {
             require!(
-                (change_amount < 0 && native_after_change >= 0)
-                    || (change_amount > 0 && native_after_change < 1),
+                native_after_change < 1 || native_after_change <= native,
```
ckamm, line 348: Hmm.
Four cases:
- balance was negative, stayed <1: allow, ok
- balance was negative, became >=1: forbid, ok
- balance was positive, increased: forbid, ok
- balance was positive, decreased: allow, ok
seems good

programs/mango-v4/src/instructions/flash_loan.rs
```diff
@@ -343,10 +343,15 @@ pub fn flash_loan_end<'key, 'accounts, 'remaining, 'info>(
...
+                MangoError::TokenInReduceOnlyMode
+            );
+        }
+        if bank.are_borrows_reduce_only() {
+            require!(
+                native_after_change >= native || native_after_change >= 0,
```
ckamm, line 354: again looking at the four cases
- balance was negative, decreased: forbid, ok
- balance was negative, increased: allow, ok
- balance was positive, stayed >=0: allow, ok
- balance was positive, went < 0: forbid, ok

programs/mango-v4/src/instructions/flash_loan.rs
```diff
@@ -343,10 +343,15 @@ pub fn flash_loan_end<'key, 'accounts, 'remaining, 'info>(
 
         let change_amount = change.amount - loan_origination_fee;
         let native_after_change = native + change_amount;
-        if bank.is_reduce_only() {
+        if bank.are_deposits_reduce_only() {
```
ckamm, line 346: minor: for whatever reason a function starting with `are_` feels unusual, weird

programs/mango-v4/src/instructions/token_force_close_borrows_with_token.rs
```diff
@@ -0,0 +1,202 @@
...
+use crate::logs::TokenBalanceLog;
+use crate::state::*;
+use anchor_lang::prelude::*;
+use fixed::types::I80F48;
+
+pub fn token_force_close_borrows_with_token(
+    ctx: Context<TokenForceCloseBorrowsWithToken>,
+    // which asset tokens are allowed, is checked at # 3
+    asset_token_index: TokenIndex,
+    // token's force_close flag is checked at # 2
```
ckamm, line 13: minor, can you use `#2` and `#3`, so it's the same string as below? I put a space there only so github doesn't make it a link to issue # 2

programs/mango-v4/tests/cases/test_force_close.rs
```diff
@@ -0,0 +1,271 @@
...
+        owner,
+        99,
+        &context.users[1],
+        mints,
+        100000,
+        0,
+    )
+    .await;
+
+    let liqor = send_tx(
```
ckamm, line 58: the liqor and liqee could also be created with create_funded_account more succinctly, I think?

programs/mango-v4/src/instructions/token_edit.rs
```diff
@@ -36,8 +36,9 @@ pub fn token_edit(
     deposit_weight_scale_start_quote_opt: Option<f64>,
     reset_stable_price: bool,
     reset_net_borrow_limit: bool,
-    reduce_only_opt: Option<bool>,
+    reduce_only_opt: Option<u8>,
     name_opt: Option<String>,
+    force_close_opt: Option<bool>,
```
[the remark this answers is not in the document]
microwavedcola1: yeah, what do you mean by "collect"? collect what?

programs/mango-v4/src/instructions/flash_loan.rs
microwavedcola1: how would you name it? I use is for singular and are for plural usually

programs/mango-v4/src/instructions/token_edit.rs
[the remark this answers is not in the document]
ckamm: If we wanted to parse instructions from the recorded transactions on-chain that'd be difficult because there are multiple versions of the same instruction with different arguments lists.

ckamm (approved):

programs/mango-v4/src/instructions/token_liq_with_token.rs
```diff
@@ -124,7 +124,7 @@ pub(crate) fn liquidation_action(
     //   assets = liabs * liab_oracle_price / asset_oracle_price * fee_factor
     //   assets = liabs * liab_oracle_price_adjusted / asset_oracle_price
     //          = liabs * lopa / aop
-    let fee_factor = I80F48::ONE + asset_bank.liquidation_fee + liab_bank.liquidation_fee;
+    let fee_factor = I80F48::ONE + liab_bank.liquidation_fee;
```
ckamm, line 127: let's move this to a separate PR and update the doc comment

PR #519, opened 2023-03-26
add leverage calculation

mschneider:


ts/client/src/accounts/mangoAccount.ts
```diff
@@ -390,6 +390,18 @@ export class MangoAccount {
...
+  /**
+   * Ratio of liabilities / equity
+   */
+  public getLeverage(group: Group): I80F48 {
+    const liabs = this.getAssetsValue(group);
+    const assets = this.getLiabsValue(group);
+    if (assets.isZero()) {
+      return ZERO_I80F48();
+    }
+    return liabs.div(assets.sub(liabs));
```
microwavedcola1, line 402: I think this should take into account the health type, based on that assets and liabs would be different, init for opening new positions, and maint for maintaining existing positions

PR #522, opened 2023-03-29
Fix bug where if no insurance fund position was present, ix would err…

microwavedcola1:
…or out


lib/client/src/client.rs
```diff
@@ -1013,7 +1020,7 @@ impl MangoClient {
         let settle_token_info = self.context.token(perp.market.settle_token_index);
 
         let health_remaining_ams = self
-            .derive_liquidation_health_check_remaining_account_metas(liqee.1, &[])
+            .derive_liquidation_health_check_remaining_account_metas(liqee.1, vec![], &[])
```
ckamm, line 1023: also always pass the INSURANCE_TOKEN_INDEX here

lib/client/src/context.rs
```diff
@@ -280,12 +280,22 @@ impl MangoGroupContext {
...
         let mut oracles = vec![];
 
+        let mut account = account1.clone();
+        for affected_token_index in &affected_tokens {
+            account.ensure_token_position(affected_token_index)?;
+        }
+        let mut account = account2.clone();
+        for affected_token_index in &affected_tokens {
+            account.ensure_token_position(*affected_token_index)?;
+        }
```
ckamm, line 297: It doesn't sound like this would do anything since the new `account` isn't used anywhere? Instead, do something like `.chain(affected_tokens.iter())` before the `unique()` a few lines down

lib/client/src/context.rs
```diff
@@ -280,12 +280,22 @@ impl MangoGroupContext {
         &self,
         account1: &MangoAccountValue,
         account2: &MangoAccountValue,
+        affected_tokens: Vec<TokenIndex>,
```
ckamm, line 283: prefer passing a `&[TokenIndex]` slice - no need to pass in ownership

ckamm (approved):

ckamm:
It's weird how the rust format ci job compains about conflict separators that don't show up in the diff on github. Please double check before merging.

PR #523, opened 2023-03-29
sync rate params to latest proposal

microwavedcola1:
image


programs/mango-v4/src/instructions/token_register_trustless.rs
```diff
@@ -45,10 +45,10 @@ pub fn token_register_trustless(
...
         adjustment_factor: I80F48::from_num(0.004),
-        util0: I80F48::from_num(0.7),
-        rate0: I80F48::from_num(0.1),
-        util1: I80F48::from_num(0.85),
-        rate1: I80F48::from_num(0.2),
+        util0: I80F48::from_num(0.5),
+        rate0: I80F48::from_num(0.018),
+        util1: I80F48::from_num(0.8),
+        rate1: I80F48::from_num(0.05),
         max_rate: I80F48::from_num(2.0),
```
ckamm, line 52: This is not quite right. Let's start it above the minimum interest rate. Proposal: rate0: 4*0.018 = 0.072, rate1: 4*0.05 = 0.2.

Note how the recent proposals also set max_rate to 0.5. We're not doing this here, so the other rates also need to be 4x the size.

ckamm (approved):

PR #525, opened 2023-03-30
Perp force close positions in a market

microwavedcola1:


programs/mango-v4/src/instructions/perp_force_close_position.rs
```diff
@@ -0,0 +1,49 @@
...
+        .unwrap();
+    let oracle_price = perp_market
+        .oracle_price(
+            &AccountInfoRef::borrow(ctx.accounts.oracle.as_ref())?,
+            Some(staleness_slot),
+        )
+        .unwrap();
+    let quote_transfer = I80F48::from(base_transfer * perp_market.base_lot_size) * oracle_price;
+
+    account_a_perp_position.record_trade(&mut perp_market, base_transfer, -quote_transfer);
```
ckamm, line 45: This would do the wrong thing if account_a's base position is positive. Let's require account_a.base_pos() > 0 and the other < 0, then we don't need the signum check and abs() use when computing base_transfer.

programs/mango-v4/src/instructions/perp_force_close_position.rs
```diff
@@ -0,0 +1,45 @@
...
+        MangoError::SomeError
+    );
+
+    let base_transfer = account_a_perp_position
+        .base_position_lots()
+        .min(account_b_perp_position.base_position_lots().abs())
+        .max(0);
+
+    let oracle_price = perp_market
+        .oracle_price(&AccountInfoRef::borrow(ctx.accounts.oracle.as_ref())?, None)
```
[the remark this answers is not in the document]
ckamm: Hmm. Your thinking is that we want this to work even if the oracle isn't reliable? That makes some sense, though force-closing at an outdated price would also be a problem. I'd say we _do_ the staleness check, and if we don't want it, the PerpMarketEdit that sets force_close=1 could also disable the staleness check.

programs/mango-v4/src/instructions/perp_force_close_position.rs
```diff
@@ -0,0 +1,45 @@
...
+        .unwrap();
+    let quote_transfer = I80F48::from(base_transfer * perp_market.base_lot_size) * oracle_price;
+
+    account_a_perp_position.record_trade(&mut perp_market, -base_transfer, quote_transfer);
+    account_b_perp_position.record_trade(&mut perp_market, base_transfer, -quote_transfer);
```
ckamm, line 42: TODO: emit perp balance logs and probably a force-close trade log

ckamm (approved):

PR #534, opened 2023-04-10
Net borrow limit: Separate out tracking and checking

ckamm:
That way it's easier to be specific about where the limit should be checked.


programs/mango-v4/src/instructions/token_withdraw.rs
```diff
@@ -69,8 +69,8 @@ pub fn token_withdraw(ctx: Context<TokenWithdraw>, amount: u64, allow_borrow: bo
         position,
         amount_i80f48,
         Clock::get()?.unix_timestamp.try_into().unwrap(),
-        oracle_price,
     )?;
+    bank.check_net_borrows(oracle_price)?;
```
ckamm, line 73: Only do this if `is_borrow`!

microwavedcola1 (approved):

PR #536, opened 2023-04-13
Don't incentivise using asset tokens with high liquidation fee during liquidation

microwavedcola1:
> It might lead to bad incentives to sum up the liq fees like this. Example: Liquidating SOL with USDC gives liqors a 5% liq fee, but liquidating SOL with USDT gives them 7.5%...

### DIFF
--- a/programs/mango-v4/tests/cases/test_liq_tokens.rs
+++ b/programs/mango-v4/tests/cases/test_liq_tokens.rs
@@ -331,10 +331,10 @@ async fn test_liq_tokens_with_token() -> Result<(), TransportError> {
     .await
     .unwrap();
 
-    // the we only have 20 collateral2, and can trade them for 20 / 1.04 = 19.2 borrow2
+    // the we only have 20 collateral2, and can trade them for 20 / 1.02 = 19.6 borrow2
     assert_eq!(
         account_position(solana, account, borrow_token2.bank).await,
-        -50 + 19
+        -50 + 20
     );
     assert!(account_position_closed(solana, account, collateral_token2.bank).await,);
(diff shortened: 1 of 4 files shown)


ckamm (approved):

ckamm:
The PR name about non-usdc tokens is a bit misleading. "Don't incentivise using asset tokens with high liquidation fee during liquidation"?

ckamm:
Hmm, what happens if we're liquidating a USDC borrow? Let's discuss.

programs/mango-v4/tests/cases/test_liq_tokens.rs
```diff
@@ -474,12 +474,10 @@ async fn test_liq_tokens_with_token() -> Result<(), TransportError> {
...
     .unwrap();
 
-    // Liqee's remaining collateral got dusted, only borrows remain
-    // but the borrow amount is so tiny, that being_liquidated is already switched off
     let liqee = get_mango_account(solana, account).await;
     assert_eq!(liqee.active_token_positions().count(), 1);
-    assert!(account_position_f64(solana, account, borrow_token1.bank).await > -1.0);
-    assert!(!liqee.being_liquidated());
+    assert!(account_position_f64(solana, account, borrow_token1.bank).await > -2.74);
+    assert!(liqee.being_liquidated());
```
ckamm, line 480: checking that being_liquidated turns off in the end is part of the test, I'll see if I can adjust this

ckamm: On reading the code: it's fine, bankruptcy is handled in another test. I just re-added the comment

PR #541, opened 2023-04-14
Spot based token rebalancer

microwavedcola1:


ts/client/scripts/rebalancer.ts
```diff
@@ -0,0 +1,175 @@
...
+      console.log(`- Rebalancing ${tokenBalance} $${baseBank.name}`);
+
+      const price = await serum3Market.computePriceForMarketOrderOfSize(
+        client,
+        group,
+        Math.abs(tokenBalance),
+        tokenBalance > 0 ? 'sell' : 'buy',
+      );
+      try {
+        const sig = await client.serum3PlaceOrder(
```
microwavedcola1, line 124: closing oo in same tx, gives this https://explorer.solana.com/tx/2y9tA426pRC28yxqoZSFFfrhSvRoo1UhUPRSZkr7R5ZZJMVeruJfKkSHQN2Q9kpUNLaxYc3Tbm5f5MZG4NUV4F6U, not sure what is wrong, now I have a separate bulk closure of all oo accounts below separately which I confirmed to work

ts/client/scripts/rebalancer.ts
```diff
@@ -0,0 +1,175 @@
...
+          group,
+          mangoAccount,
+          serum3Market.serumMarketExternal,
+          tokenBalance > 0 ? Serum3Side.ask : Serum3Side.bid,
+          price,
+          Math.abs(tokenBalance),
+          Serum3SelfTradeBehavior.decrementTake,
+          Serum3OrderType.immediateOrCancel,
```
mschneider, line 132: better to use limit order instead of IOC

ts/client/scripts/rebalancer.ts
```diff
@@ -0,0 +1,175 @@
...
+      // Skip if quantity is too small
+      if (maxBaseQuantity.eq(new BN(0))) {
+        console.log(
+          ` - Not rebalancing ${tokenBalance} $${baseBank.name}, quantity too small`,
+        );
+        continue;
+      }
+      console.log(`- Rebalancing ${tokenBalance} $${baseBank.name}`);
+
+      const price = await serum3Market.computePriceForMarketOrderOfSize(
```
mschneider, line 117: price should be oracle - liquidation fee

ts/client/scripts/rebalancer.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    mangoAccount = await mangoAccount.reload(client);
+    const ixs: TransactionInstruction[] = [];
+    for (const serum3OoMarketIndex of Array.from(
+      mangoAccount.serum3OosMapByMarketIndex.keys(),
+    )) {
+      const serum3ExternalPk = group.serum3MarketsMapByMarketIndex.get(
+        serum3OoMarketIndex as MarketIndex,
+      )!.serumMarketExternal;
+      // 12502 cu per market
```
mschneider, line 150: cancel, settle, close

ts/client/scripts/rebalancer.ts
```diff
@@ -0,0 +1,175 @@
...
+        tokenBalance > 0 ? 'sell' : 'buy',
+      );
+      try {
+        const sig = await client.serum3PlaceOrder(
+          group,
+          mangoAccount,
+          serum3Market.serumMarketExternal,
+          tokenBalance > 0 ? Serum3Side.ask : Serum3Side.bid,
+          price,
+          Math.abs(tokenBalance),
```
mschneider, line 130: might want to round up tokenBalance to next minOrderSize for the bid case

bin/liquidator/src/rebalance.rs
```diff
@@ -77,7 +77,7 @@ impl Rebalancer {
     pub async fn zero_all_non_quote(&self) -> anyhow::Result<()> {
         log::trace!("checking for rebalance: {}", self.mango_account_address);
 
-        self.rebalance_tokens().await?;
+        // self.rebalance_tokens().await?;
```
ckamm, line 80: don't merge this anymore

ts/client/scripts/rebalancer.ts
```diff
@@ -0,0 +1,197 @@
...
+      }
+      const serum3Market = serum3Markets[0];
+      const serum3MarketExternal = group.serum3ExternalMarketsMap.get(
+        serum3Market.serumMarketExternal.toBase58(),
+      )!;
+      const maxBaseQuantity = serum3MarketExternal.baseSizeNumberToLots(
+        Math.abs(tokenBalance),
+      );
+      // Skip if quantity is too small
+      if (maxBaseQuantity.eq(new BN(0))) {
```
ckamm, line 109: too small means ==0? probably outdated comments

ts/client/scripts/rebalancer.ts
```diff
@@ -0,0 +1,197 @@
...
+          [
+            ...(await client.serum3PlaceOrderIx(
+              group,
+              mangoAccount,
+              serum3Market.serumMarketExternal,
+              tokenBalance > 0 ? Serum3Side.ask : Serum3Side.bid,
+              price,
+              Math.abs(tokenBalance),
+              Serum3SelfTradeBehavior.decrementTake,
+              Serum3OrderType.limit,
```
ckamm, line 134: why use limit + cancel all instead of ImmediateOrCancel?

ts/client/src/accounts/serum3.ts
```diff
@@ -140,6 +140,40 @@ export class Serum3Market {
     );
   }
 
+  public async computePriceForMarketOrderOfSize(
```
ckamm, line 143: is this unused?

ts/client/src/client.ts
```diff
@@ -1601,24 +1601,42 @@ export class MangoClient {
...
       externalMarketPk,
     );
 
-    return await this.sendAndConfirmTransactionForGroup(group, [
-      ...placeOrderIxes,
-      settleIx,
-    ]);
+    const ixs = [...placeOrderIxes, settleIx];
+
+    if (closeOpenOrdersAccount) {
```
ckamm, line 1613: is this still a useful flag, given that we saw that when you actually take you'll not be able to close without processing some events? (at least that's what I heard)

ts/client/src/accounts/serum3.ts
microwavedcola1: unused in this PR, but useful in general, max suggested to not market order but use a limit order with price which is oracle price + liquidation fee. I'd like to keep this function if ok

ts/client/scripts/rebalancer.ts
microwavedcola1: no, basically if the quantity is too small for the lot size on the openbook market, at least this is what I am trying to achieve

ts/client/scripts/rebalancer.ts
microwavedcola1: this is what max suggested

> better to use limit order instead of IOC, you have a higher chance of finding liquidity if you stay on the book for a few slots (arb bots)

> price should be oracle - liquidation fee. you're trying to catch a falling knife, looking at the orderbook will add latency, and the price will have moved most likely by the time you send your order

ts/client/src/client.ts
microwavedcola1: no, removed now

ts/client/scripts/rebalancer.ts
microwavedcola1: wdyt?

ts/client/scripts/rebalancer.ts
ckamm: ah!

ts/client/scripts/rebalancer.ts
ckamm: Yeah, makes sense. But if we cancel immediately in the same tx we're not resting the limit order on the book for a few slots. The current limit + cancel all in one tx should be effectively a costly immediateorcancel?

But we'd need to change this a good bit to deal with long-lived orders, so that sounds like something for a future PR

ts/client/src/accounts/serum3.ts
ckamm: ok, it also has some random constants though...

ckamm (approved):

ts/client/scripts/rebalancer.ts
microwavedcola1: agreed, I switched back to a immediateorcancel for now

PR #542, opened 2023-04-14
FlashLoan: Don't deduce vault_len, add more checks

ckamm:
It looks like it wasn't possible to exploit the program by re-initializing the user-owned token accounts used in flash loan because the later use of health_ais with the health computation would error out if any token account was included there.

However, the check and a few others were too indirect. In this patch:
- We explicitly pass the number of loans into FlashLoanEnd (verified from FlashLoanBegin)
- Add explicit checks for token mints, so it's no longer possible to use token accounts for forein mints in Begin when the loan amount is zero, and it's clearer to see that the bookkeeping in End won't break if the user reinited the account for a different mint.
- Also add a few other extra comments and checks.

The updated FlashLoanEnd instruction is called FlashLoanEndV2

### DIFF
--- a/lib/client/src/client.rs
+++ b/lib/client/src/client.rs
@@ -1293,6 +1293,7 @@ impl MangoClient {
             },
             0u64,
         ];
+        let num_loans: u8 = loan_amounts.len().try_into().unwrap();
 
         // This relies on the fact that health account banks will be identical to the first_bank above!
         let health_ams = self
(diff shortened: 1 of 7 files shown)


ckamm:
Hm, I guess it'd be preferable to keep old instructions in the idl, I'll do that.

microwavedcola1 (approved):

PR #546, opened 2023-04-17
Ts keeper (backup for rust keeper)

microwavedcola1:


ts/client/scripts/keeper/keeper.ts
```diff
@@ -0,0 +1,211 @@
...
+  const mangoAccount = await client.getMangoAccount(
+    new PublicKey(MANGO_ACCOUNT_PK),
+  );
+  const group = await client.getGroup(mangoAccount.group);
+  await group.reloadAll(client);
+
+  updateBanks(client, group);
+  consumeEvents(client, group);
+  updateFunding(client, group);
+  checkNewListingsAndAbort(client, group);
```
ckamm, line 208: does this work? you don't need to await the promises returned here?

ckamm (approved):

riordanp (approved):

ts/client/scripts/keeper/keeper.ts
microwavedcola1: yes it works, we dont await the promises, this way we can basically trigger all functions

microwavedcola1: already deployed at `heroku logs -a mango-v4-ts-keeper-1 -t`

PR #548, opened 2023-04-17
wip: market maker improvements

microwavedcola1:


ts/client/scripts/mm/params/default.json
```diff
@@ -36,7 +42,10 @@
         "takeSpammers": true,
         "spammerCharge": 2,
         "charge": 0.002,
-        "krakenCode": "XETHZUSD"
+        "krakenCode": "XETHZUSD",
+        "binanceCode": "ETHUSDT",
```
mschneider, line 46: ETHUSDT is not ETHUSD unfortunately, you probably need to convert this price using BINANCE:USDCUSDT or  BITFINEX:USDTUSD

PR #550, opened 2023-04-19
Configurable perp settle token

ckamm:


programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
```diff
@@ -230,154 +238,283 @@ pub(crate) fn liquidation_action(
...
+
     // Take over the liqee's base in exchange for quote
     let liqee_base_lots = liqee_perp_position.base_position_lots();
-    // Each lot the base position gets closer to 0, the unweighted perp health contribution
-    // increases by this amount.
-    let unweighted_health_per_lot;
+
+    // Each lot the base position gets closer to 0, the "unweighted health unsettled pnl"
+    // increases by this amount
+    let quote_per_lot;
```
ckamm, line 275: uhupnl_per_lot sounds more appropriate. what does "quote" even mean here?

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
```diff
@@ -230,154 +238,283 @@ pub(crate) fn liquidation_action(
...
-        // How much are we willing to increase the unweighted perp health?
-        let health_limit = health_amount
-            .min(-current_expected_health)
-            .max(I80F48::ZERO);
+                           mut uhupnl_limit: I80F48,
+                           // How much effective settle token will be gained when taking future
+                           // upnl settlement into account.
+                           expected_quote_per_lot: I80F48,
+                           // Effective settle token gained directly from the operation (before settlement)
+                           actual_quote_per_lot: I80F48,
```
ckamm, line 333: calling this quote is confusing

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
```diff
@@ -230,154 +238,283 @@ pub(crate) fn liquidation_action(
...
+    // Helper function to reduce base position in exchange for effective settle token balance changes
     let mut reduce_base = |step: &str,
-                           health_amount: I80F48,
-                           health_per_lot: I80F48,
-                           current_unweighted_perp_health: &mut I80F48| {
-        // How much are we willing to increase the unweighted perp health?
-        let health_limit = health_amount
-            .min(-current_expected_health)
-            .max(I80F48::ZERO);
+                           mut uhupnl_limit: I80F48,
```
ckamm, line 328: looks like this doesn't need to be `mut`

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
```diff
@@ -230,154 +238,283 @@ pub(crate) fn liquidation_action(
...
+                           // upnl settlement into account.
+                           expected_quote_per_lot: I80F48,
+                           // Effective settle token gained directly from the operation (before settlement)
+                           actual_quote_per_lot: I80F48,
+                           uhupnl_per_lot: I80F48,
+                           current_uhupnl: &mut I80F48,
+                           current_settle_token: &mut I80F48,
+                           current_health: &mut I80F48| {
+        let expected_settle_token = *current_settle_token;
+        let expected_health = *current_health;
```
ckamm, line 339: these look unnecessary

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
```diff
@@ -230,154 +238,283 @@ pub(crate) fn liquidation_action(
...
+            &mut current_uhupnl,
+            &mut current_settle_token,
+            &mut current_health,
         );
     }
 
     //
     // Step 2: If perp unsettled health is positive but below max_settle, perp base position reductions
     // benefit account health slightly less because of the settlement liquidation fee.
     //
```
ckamm, line 474: This "step" number doesn't match up with the numbering above

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
```diff
@@ -386,7 +523,7 @@ pub(crate) fn liquidation_action(
     // liqee and liqors entry and break even prices.
     //
     let base_transfer = direction * base_reduction;
-    let quote_transfer = -I80F48::from(base_transfer) * oracle_price_per_lot * fee_factor;
+    let quote_transfer = -I80F48::from(base_transfer) * oracle_price_per_lot * base_fee_factor;
```
ckamm, line 526: is there a way to double check that these numbers are sane?

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
```diff
@@ -436,34 +561,33 @@ pub(crate) fn liquidation_action(
...
-        }
+        liqor_perp_position.record_liquidation_pnl_takeover(pnl_transfer, limit_transfer);
+        liqee_perp_position.record_settle(pnl_transfer);
+
+        // Update the accounts' perp_spot_transfer statistics.
+        let transfer_i64 = token_transfer.round_to_zero().to_num::<i64>();
+        liqor_perp_position.perp_spot_transfers -= transfer_i64;
+        liqee_perp_position.perp_spot_transfers += transfer_i64;
+        liqor.fixed.perp_spot_transfers -= transfer_i64;
+        liqee.fixed.perp_spot_transfers += transfer_i64;
```
ckamm, line 572: the direction of perp_spot_transfers can always use double-checking

ckamm (commented):
still missing "negative_perp_or_bankruptcy"

bin/liquidator/src/liquidate.rs
```diff
@@ -541,7 +464,7 @@ impl<'a> LiquidateHelper<'a> {
     }
 
     async fn token_liq_bankruptcy(&self) -> anyhow::Result<Option<Signature>> {
-        if !self.health_cache.in_phase3_liquidation() || !self.health_cache.has_spot_borrows() {
+        if !self.health_cache.in_phase3_liquidation() || !self.health_cache.has_liq_spot_borrows() {
```
microwavedcola1, line 467: nit: has_spot_borrows was more appropriate imo then has_liq_spot_borrows, it still only checks if there is any token with borrow, nothing more

programs/mango-v4/src/health/cache.rs
```diff
@@ -641,19 +890,32 @@ impl HealthCache {
...
-        for perp_info in self.perp_infos.iter() {
-            let contrib = perp_info.health_contribution(health_type);
-            action(contrib);
-        }
     }
 
-    /// Compute the health when it comes to settling perp pnl
+    /// Returns how much pnl is settleable for a given settle token.
+    ///
+    /// The idea of this limit is that settlement is only permissible as long as there are
```
microwavedcola1, line 903: nit:
> it should not be allowed to settle that perp loss into the spot world fully
could we expand on why?

lib/client/src/perp_pnl.rs
```diff
@@ -95,11 +95,11 @@ pub async fn fetch_top(
             } else {
                 I80F48::ZERO
             };
-            let perp_settle_health = crate::health_cache::new(context, account_fetcher, &acc)
+            let perp_max_settle = crate::health_cache::new(context, account_fetcher, &acc)
                 .await?
-                .perp_settle_health();
-            let settleable_pnl = if perp_settle_health > 0 {
-                (*pnl).max(-perp_settle_health)
+                .perp_max_settle(perp_market.settle_token_index)?;
```
microwavedcola1, line 100: note to self: port to ts

programs/mango-v4/src/accounts_ix/token_liq_bankruptcy.rs
```diff
@@ -40,6 +40,7 @@ pub struct TokenLiqBankruptcy<'info> {
 
     #[account(mut)]
     // address is checked at #2 a) and b)
+    // better name would be "insurance_bank_vault"
```
microwavedcola1, line 43: imo, we should do announcements and remove unused ixs from program after newer versions are 2-3 releases behind to slim down idl

programs/mango-v4/src/health/cache.rs
```diff
@@ -89,6 +89,51 @@ pub fn compute_health(
...
+
+/// How much of a token can be gained before health increases to zero?
+pub fn spot_amount_given_for_health_zero(
+    health: I80F48,
+    starting_spot: I80F48,
+    asset_weighted_price: I80F48,
+    liab_weighted_price: I80F48,
+) -> Result<I80F48> {
+    // asset/liab prices are reversed intentionally
+    spot_amount_taken_for_health_zero(
```
microwavedcola1, line 129: nit: assert for health lower than 0?

programs/mango-v4/src/health/cache.rs
```diff
@@ -89,6 +89,51 @@ pub fn compute_health(
...
+        -health,
+        -starting_spot,
```
microwavedcola1, line 131: nit: can you please explain to me why this (negating the current position balance) works?

programs/mango-v4/src/health/cache.rs
```diff
@@ -351,35 +441,135 @@ pub struct HealthCache {
...
+        for token_info in self.token_infos.iter() {
+            // For each token, health only considers the effective token position. But for
+            // this function we want to distinguish the contribution from token deposits from
+            // contributions by perp markets.
+            // However, the overall weight is determined by the sum, so first collect all
+            // assets parts and all liab parts and then determine the actual values.
+            let mut asset_balance = I80F48::ZERO;
+            let mut liab_balance = I80F48::ZERO;
+
+            add(
```
microwavedcola1, line 508: 508-519, what is the difference between adding balance_spot + hupnl vs simply using balance_spot_and_perp

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
```diff
@@ -230,163 +238,294 @@ pub(crate) fn liquidation_action(
...
-        let quote_init_liab_weight = I80F48::ONE;
+        // health gets increased by `base * base_price * perp_init_liab_weight`
+        // and reduced by `base * base_price * (1 + liq_fee)`
         direction = 1;
-        fee_factor = I80F48::ONE + perp_market.base_liquidation_fee;
-        let liab_price = perp_info.prices.liab(HealthType::LiquidationEnd);
-        unweighted_health_per_lot = liab_price * base_lot_size * perp_market.init_base_liab_weight
-            - oracle_price_per_lot * quote_init_liab_weight * fee_factor;
+        base_fee_factor = I80F48::ONE + perp_market.base_liquidation_fee;
+        uhupnl_per_lot =
```
microwavedcola1, line 306: liab_price * base_lot_size is getting simplified to oracle_price_per_lot, is this intentional? same in the else clause

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
```diff
@@ -230,163 +238,294 @@ pub(crate) fn liquidation_action(
...
             .min(max_base_transfer.abs() - base_reduction)
             .max(0);
-        let unweighted_change = I80F48::from(base_lots) * unweighted_health_per_lot;
-        let current_unweighted = *current_unweighted_perp_health;
-        let new_unweighted_perp = current_unweighted + unweighted_change;
-        let new_expected_perp = expected_perp_health(new_unweighted_perp);
-        let new_expected_health =
-            current_expected_health + (new_expected_perp - current_expected_perp_health);
+
+        let expected_settle_token_gain = expected_settle_token_per_lot * I80F48::from(base_lots);
```
microwavedcola1, line 356: are the expected_* variables just for logging?

programs/mango-v4/src/instructions/perp_liq_negative_pnl_or_bankruptcy.rs
```diff
@@ -129,46 +283,76 @@ pub fn perp_liq_negative_pnl_or_bankruptcy(
...
-        let max_liab_transfer_from_liqee =
-            (-liqee_pnl).min(-liqee_liq_end_health).max(I80F48::ZERO);
-        let liab_transfer = max_liab_transfer_from_liqee
+        // recompute for safety
+        liqee_pnl = liqee_perp_position.unsettled_pnl(&perp_market, perp_oracle_price)?;
+
+        // Each unit of pnl increase (towards 0) increases health, but the amount depends on whether
+        // the health token position is negative or positive.
+        // Compute how much pnl would need to be increased to reach liq end health 0 (while ignoring
+        // liqee_pnl and other constraints for now)
```
microwavedcola1, line 323: can the constraints be freely ignored?

programs/mango-v4/src/instructions/token_liq_bankruptcy.rs
```diff
@@ -59,7 +59,18 @@ pub fn token_liq_bankruptcy(
...
+        liqee_health_cache.effective_token_balances(HealthType::LiquidationEnd);
+    let liqee_liab_health_balance = liqee_health_token_balances
+        [liqee_health_cache.token_info_index(liab_token_index)?]
+    .spot_and_perp;
+
+    // Allow token bankruptcy only while the spot position and health token position are both negative.
+    // In particular, a very negative perp hupnl does not allow token bankruptcy to happen,
+    // and if the perp hupnl is positive, we need to liquidate that before dealing with token
+    // bankruptcy!
+    let mut remaining_liab_loss = (-initial_liab_native).min(-liqee_liab_health_balance);
```
microwavedcola1, line 73: hmm in case of a very negative hupnl, -liqee_liab_health_balance > -initial_liab_native, that means ix would still proceed, contrary to the comment

bin/liquidator/src/liquidate.rs
ckamm: No, it's different now. It checks if there is a spot balance and the sum of the spot balance and all perp uhupnl (which is weighted!) is negative.

programs/mango-v4/src/accounts_ix/token_liq_bankruptcy.rs
ckamm: Keeping them in the idl allows the explorer to still work nicely when browsing old transaction, that's why I would prefer to keep everything in the idl. But yeah, we do need to figure out growing the idl account above 10k soonish.

programs/mango-v4/src/health/cache.rs
ckamm: Increasing health by depositing spot is equivalent to `spot_amount_taken_for_health_zero()`. You want to give spot and the weights change depending on the balance. 
- If spot balance is negative, use the liab weight. (equivalent to spot_amount_taken_for_health_zero() taking a positive balance with the asset weight)
- If spot balance is positive, use the asset weight. (equivalent to the other function taking a negative balance with the liab weight)

programs/mango-v4/src/health/cache.rs
```diff
@@ -351,35 +441,135 @@ pub struct HealthCache {
...
+    ///
+    /// Same example if liabs are held stable:
+    /// - liabs: $12 * SOL_liab_weight
+    /// - assets: $10 * SOL_liab_weight
+    ///
+    /// The value `assets - liabs` is the health and the same in both cases.
+    fn health_assets_and_liabs(
+        &self,
+        health_type: HealthType,
+        stable_assets: bool,
```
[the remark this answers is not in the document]
ckamm: Yep, this needs discussion, I'm not sure the current decision is good.

programs/mango-v4/src/health/cache.rs
ckamm: It's a key difference! In this code we allocate the token balance and the perp hupnl separately to assets and liabs. If we only allocated the sum, the result would be way different. Imagine having $5 spot and $-5 SOL-PERP hupnl. This should be accounted as $5 in assets and -$5 in liabs (health ratio 50%), not as 0 (health ratio inf).

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
ckamm: Yes. It used to be different. I'm not sure it's worth keeping them for logging purposes. What do you think?

programs/mango-v4/src/instructions/token_liq_bankruptcy.rs
ckamm: If liab spot or liab balance is positive, the require() will be hit.

If both are negative and liab balance is more negative, then remaining_liab_loss will be -liab_spot. So we only liquidate the actual spot borrows.

Would it be helpful to write `-(initial_liab_native.max(liqee_liab_health_balance))` instead?

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
microwavedcola1: honestly, I need to look at some real world liq logs to see how the logging in this ix is, so lets revisit once we have some real world examples

ckamm: ok

programs/mango-v4/src/instructions/perp_liq_negative_pnl_or_bankruptcy.rs
ckamm: they are just ignored for max_for_health. The two lines below then limit to liqee_pnl and max_liab_transfer

programs/mango-v4/src/health/cache.rs
```diff
@@ -377,35 +471,135 @@ pub struct HealthCache {
...
-        let sum = |contrib| {
-            if contrib > 0 {
-                assets += contrib;
+    pub fn health_assets_and_liabs_stable_assets(
+        &self,
+        health_type: HealthType,
+    ) -> (I80F48, I80F48) {
+        self.health_assets_and_liabs(health_type, true)
+    }
+    pub fn health_assets_and_liabs_stable_liabs(
```
microwavedcola1, line 489: hmm this seems to be only used in the group level deposit limit feature which is disable for now, I was wondering why these family of functions are in program, anyway I guess we can keep them here

programs/mango-v4/src/health/cache.rs
```diff
@@ -648,9 +859,54 @@ impl HealthCache {
...
+
+    /// Implementation of effective_token_balances()
+    ///
+    /// The ignore_negative_perp flag exists for perp_max_settle(). When it is enabled, all negative
+    /// token contributions from perp markets are ignored. That's useful for knowing how much token
+    /// collateral is available when limiting negative upnl settlement.
+    fn effective_token_balances_internal(
+        &self,
+        health_type: HealthType,
+        ignore_negative_perp: bool,
```
microwavedcola1, line 882: I have now pushed a change in client where this field name now matches

programs/mango-v4/src/instructions/perp_liq_base_or_positive_pnl.rs
```diff
@@ -230,163 +238,295 @@ pub(crate) fn liquidation_action(
...
+        let new_expected_settle_token = *current_settle_token + expected_settle_token_gain;
+        let new_expected_health = *current_health
+            - settle_token_info.health_contribution(liq_end_type, *current_settle_token)
+            + settle_token_info.health_contribution(liq_end_type, new_expected_settle_token);
+
+        let uhupnl_gain = uhupnl_per_lot * I80F48::from(base_lots);
+        let new_uhupnl = *current_uhupnl + uhupnl_gain;
+
+        let new_settle_token =
+            *current_settle_token + actual_settle_token_per_lot * I80F48::from(base_lots);
```
microwavedcola1, line 367: hmm shouldn't the base_lots be recalculated using actual_settle_token_per_lot?

microwavedcola1 (approved):

PR #551, opened 2023-04-19
Force close for serum3 market

microwavedcola1:


programs/mango-v4/src/instructions/serum3_edit_market.rs
```diff
@@ -24,6 +25,18 @@ pub fn serum3_edit_market(
...
 
+    if let Some(force_close) = force_close_opt {
+        msg!(
+            "Force close: old - {:?}, new - {:?}",
+            serum3_market.force_close,
+            u8::from(force_close)
+        );
+        serum3_market.force_close = u8::from(force_close);
+        if !force_close {
+            require_group_admin = true;
```
ckamm, line 36: let's always require the group admin

programs/mango-v4/src/instructions/serum3_liq_force_cancel_orders.rs
```diff
@@ -63,7 +63,7 @@ pub fn serum3_liq_force_cancel_orders(
         {
             let result = account.check_liquidatable(&health_cache);
             if account.fixed.is_operational() {
-                if !result? {
+                if !result? && !serum_market.is_force_close() {
```
ckamm, line 66: This won't work: check_liquidatable() is tricky:
- returns Ok(true) when liquidatable
- returns Ok(false) if the account recovered from being_liquidated
- returns Err(HealthMustBeNegative) if account isn't liquidatable and wasn't liquidatable
- might return Err(something_else)
I think we should change it to just return a nice enum, then this code becomes less confusing too.

programs/mango-v4/src/instructions/token_liq_with_token.rs
```diff
@@ -46,7 +46,7 @@ pub fn token_liq_with_token(
     let liqee_liq_end_health = liqee_health_cache.health(HealthType::LiquidationEnd);
     liqee_health_cache.require_after_phase1_liquidation()?;
 
-    if !liqee.check_liquidatable(&liqee_health_cache)? {
+    if liqee.check_liquidatable(&liqee_health_cache)? == CheckLiquidatable::NotLiquidatable {
```
ckamm, line 49: This (and check other uses too) changes the behavior.
- Liquidatable: continue on (good)
- NotLiquidatable: previously errored the instruction, now silently returns (arguably ok, but should at least log?)
- BecameNotLiquidatable: previously early-exited the instruction to commit the being_liquidated=false change, now goes into the liquidation code and probably errors there

Suggestion: `Change this to != CL::Liquidatable` and add a log for NotLiquidatable in check_liquidatable()

programs/mango-v4/src/instructions/serum3_liq_force_cancel_orders.rs
```diff
@@ -61,17 +61,12 @@ pub fn serum3_liq_force_cancel_orders(
...
-                if !result? {
-                    return Ok(());
-                }
-            } else {
-                // Frozen accounts can always have their orders cancelled
-                if !result.is_anchor_error_with_code(MangoError::HealthMustBeNegative.into()) {
-                    // Propagate unexpected errors
-                    result?;
-                }
+            let check_liquidatable_result = account.check_liquidatable(&health_cache)?;
```
ckamm, line 64: minor: just call this `liquidatable`?

ckamm (approved):

PR #554, opened 2023-04-20
wip: various force close bots

microwavedcola1:


mschneider:
> router.ts is copied from max's mm branch

love to see it!

ts/client/scripts/router.ts
```diff
@@ -0,0 +1,410 @@
+import {
```
mschneider, line 1: this should live in the client/src so we can use it in the web frontend

ts/client/src/client.ts
```diff
@@ -1513,6 +1664,8 @@ export class MangoClient {
       );
 
     const limitPrice = serum3MarketExternal.priceNumberToLots(price);
+    console.log(size);
+    console.log(serum3MarketExternal['_decoded'].baseLotSize.toNumber());
```
ckamm, line 1668: probably leftover logging

ts/client/src/client.ts
```diff
@@ -1449,6 +1520,86 @@ export class MangoClient {
...
+    )!;
+
+    const healthRemainingAccounts: PublicKey[] =
+      this.buildHealthRemainingAccounts(
+        AccountRetriever.Fixed,
+        group,
+        [mangoAccount],
+        [
+          group.getFirstBankByTokenIndex(serum3Market.baseTokenIndex),
+          group.getFirstBankByTokenIndex(serum3Market.quoteTokenIndex),
```
ckamm, line 1543: unnecessary since base and quote positions are guaranteed to exist on the account

ts/client/src/client.ts
```diff
@@ -1449,6 +1520,86 @@ export class MangoClient {
...
+              this.program.programId,
+              mangoAccount.publicKey,
+            ),
+          ],
+        ],
+      );
+
+    console.log(
+      `${await serum3Market.findOoPda(this.programId, mangoAccount.publicKey)}`,
+    );
```
ckamm, line 1559: is this logging intentional?

ts/client/src/client.ts
```diff
@@ -1449,6 +1520,86 @@ export class MangoClient {
...
+
+    const ix = await this.program.methods
+      .serum3LiqForceCancelOrders(limit ? limit : 10)
```
ckamm, line 1562: better use `??` for these kinds of things

ts/client/src/client.ts
```diff
@@ -1449,6 +1520,86 @@ export class MangoClient {
...
+      .accounts({
+        group: group.publicKey,
+        account: mangoAccount.publicKey,
+        openOrders: await serum3Market.findOoPda(
```
ckamm, line 1566: findOoPda is called three times, maybe call once?

ts/client/src/client.ts
```diff
@@ -1280,6 +1329,28 @@ export class MangoClient {
...
+    group: Group,
+    serum3MarketIndex: MarketIndex,
+    reduceOnly: boolean,
+    forceClose: boolean,
+  ): Promise<TransactionSignature> {
+    const serum3Market =
+      group.serum3MarketsMapByMarketIndex.get(serum3MarketIndex);
+    const ix = await this.program.methods
+      .serum3EditMarket(
+        reduceOnly != null ? reduceOnly : null,
```
ckamm, line 1342: is it really allowed to pass `null` to a function expecting `boolean`?

ckamm: Also, this check is weird. The expression is different from just `reduceOnly` because so many things compare positive to `null` in js, but the intent of the code is still unclear to me. I'd rather if stuff like  reduceOnly=0 or reduceOnly=[] be an error.

microwavedcola1: the ix is expecting a Option<bool>, not just a boolean, is that what you meant?

I have now changed it to, wdyt?
image

ckamm: I don't get it. What is `reduceOnly ?? reduceOnly` supposed to do? Why not just `reduceOnly`?

microwavedcola1: ok it should be `reduceOnly ?? null`, which accounts for undefined been passed

ckamm (approved):

PR #559, opened 2023-04-24
Fix bug in closing mango account

microwavedcola1:
* Refactor code for collecting health accounts, 
* Remove scanning retriever, only build accounts for fixed retriever because they should always works, saves having to maintain 2 code blocks
* fix bug where bank oracle was skipped while closing account if same perp position in a market with same oracle was also deactivated earlier in same tx

Tested on mainnet so - via UI, created an account, deposited sol and usdc, opened perp and spot market positions, closed perp position, settled perp and spot, then called clients emptyAndCloseMangoAccount on dev and this branch, breaks on dev (https://explorer.solana.com/tx/3Wgf6GVtCYXVT3D3H1MK5TtEroKt3wYrdtnBgRNWH9JY2yvfLyov8dRUY5K7ukT8VPHJmGKovLEiLii59AV1zgNa), works on this branch (https://explorer.solana.com/tx/4Y8sVCxKxo3SwwRt1Kjc9zvk63cypbkqQd2cbCCewM1JQLGHb9jHnHhLEBvENiXnWPu34uNCYkwLrJEMgJpj57kE)

### DIFF
--- a/ts/client/src/client.ts
+++ b/ts/client/src/client.ts
@@ -719,13 +719,7 @@ export class MangoClient {
     mangoAccount: MangoAccount,
   ): Promise<TransactionSignature> {
     const healthRemainingAccounts: PublicKey[] =
-      this.buildHealthRemainingAccounts(
-        AccountRetriever.Fixed,
-        group,
-        [mangoAccount],
-        [],
-        [],
-      );
+      this.buildHealthRemainingAccounts(group, [mangoAccount], [], []);
 
     const ix = await this.program.methods
       .computeAccountData()
@@ -940,7 +934,9 @@ export class MangoClient {
     mangoAccount: MangoAccount,
   ): Promise<TransactionSignature> {
     const instructions: TransactionInstruction[] = [];
-    const healthAccountsToExclude: PublicKey[] = [];
+    const banksToExclude: Bank[] = [];
+    const perpMarketsToExclude: PerpMarket[] = [];
+    const serum3MarketsToExclude: Serum3Market[] = [];
 
     for (const serum3Account of mangoAccount.serum3Active()) {
       const serum3Market = group.serum3MarketsMapByMarketIndex.get(
@@ -952,7 +948,7 @@ export class MangoClient {
         mangoAccount,
         serum3Market.serumMarketExternal,
       );
-      healthAccountsToExclude.push(serum3Account.openOrders);
+      serum3MarketsToExclude.push(serum3Market);
       instructions.push(closeOOIx);
     }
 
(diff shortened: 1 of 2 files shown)


ts/client/src/client.ts
```diff
@@ -2979,40 +2948,43 @@ export class MangoClient {
     );
   }
 
-  public buildHealthRemainingAccounts(
```
microwavedcola1, line 2982: remove public, code is not used anywhere else in client, and in ui, if some npm client user complains, I will fix it on demand

ts/client/src/utils/rpc.ts
```diff
@@ -43,7 +43,7 @@ export async function sendTransaction(
 
   if (
     typeof payer.signTransaction === 'function' &&
-    !(payer instanceof NodeWallet)
+    !(payer instanceof NodeWallet || payer.constructor.name == 'NodeWallet')
```
microwavedcola1, line 46: This is so frustrating, but the change is minor and makes my local bot work :(

ts/client/src/client.ts
```diff
@@ -964,22 +960,21 @@ export class MangoClient {
...
-      //to withdraw from all token accounts we need to exclude previous tokens pubkeys
-      //used to build health remaining accounts
+      // To withdraw from a particular token account we need to exclude previously
+      // withdrawn tokens
       if (indexNum !== 0) {
         for (let i = indexNum; i--; i >= 0) {
           const prevToken = mangoAccount.tokensActive()[i];
           const prevBank = group.getFirstBankByTokenIndex(prevToken.tokenIndex);
-          accountsToExclude.push(prevBank.publicKey, prevBank.oracle);
+          banksToExclude.push(prevBank);
```
ckamm, line 977: Since this is in the loop over token accounts, won't this add the prevBanks again and again? Couldn't we add to banksToExclude every time a withdraw ix was made?

ts/client/src/client.ts
```diff
@@ -988,7 +983,9 @@ export class MangoClient {
         bank.mint,
         U64_MAX_BN,
         false,
-        [...accountsToExclude],
+        banksToExclude,
+        perpMarketsToExclude,
+        serum3MarketsToExclude,
```
ckamm, line 988: mmh, this probably works and I'm ok with it. However, in other places we adjust (a copy of) the mangoAccount to make the health account generation do the right thing. Might it have been easier to deactivate the perp, serum, token accounts in mangoAccount after each ix is made? Then those wouldn't need to be passed explicitly in the exclusion lists.

ts/client/src/client.ts
```diff
@@ -2979,40 +2951,43 @@ export class MangoClient {
...
-      );
-    } else {
-      return this.buildScanningAccountRetrieverHealthAccounts(
-        group,
-        mangoAccounts,
-        banks,
-        perpMarkets,
-      );
-    }
+    return this.buildFixedAccountRetrieverHealthAccounts(
```
ckamm, line 2966: this indirection could be removed now

ckamm (approved):

ckamm (approved):

PR #565, opened 2023-04-26
Risk notification bot

microwavedcola1:


ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,469 @@
...
+
+    // Sum of all assets which would be acquired in exchange for also acquiring
+    // liabs by the liqor, who would immediately want to reduce to 0
+    // Assuming assets need to be sold to USDC
+    const assets = mangoAccountsWithHealth.reduce((sum, a) => {
+      if (bank.maintAssetWeight.eq(ZERO_I80F48())) {
+        return sum.add(ZERO_I80F48());
+      }
+      // How much would health increase for every unit liab moved to liqor
+      const tokenAssetHealthContrib = ONE_I80F48()
```
microwavedcola1, line 192: not entirely sure if I am doing tokenAssetHealthContrib fully correct here..

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,469 @@
...
+              )
+          : pm.price
+              .mul(I80F48.fromU64(pm.baseLotSize))
+              .mul(pm.initBaseLiabWeight)
+              .sub(
+                I80F48.fromU64(pm.baseLotSize)
+                  .mul(pm.price)
+                  .mul(ONE_I80F48()) // quoteInitLiabWeight
+                  .mul(ONE_I80F48().add(pm.baseLiquidationFee)),
+              );
```
ckamm, line 346: these two look right! Could simplify to
`price * baselotsize * ((1+fee) - baseassetweight)` and `price * baselotsize * (baseliabweight - (1-fee))`

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,469 @@
...
+
+        const maxBaseLots = a.health
+          .min(ZERO_I80F48())
+          .abs()
+          .div(unweightedHealthPerLot.abs())
```
ckamm, line 351: again, should use the liqendhealth and divide that by that healthperlot, since that also uses init weights

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,469 @@
...
+    // who would immediately want to reduce them to 0
+    // Assuming liabs need to be bought using USDC
+    const liabs =
+      // Max liab of a particular token that would be liquidated to bring health above 0
+      mangoAccountsWithHealth.reduce((sum, a) => {
+        // How much would health increase for every unit liab moved to liqor
+        const tokenLiabHealthContrib = ONE_I80F48()
+          .sub(bank.liquidationFee)
+          .mul(bank.price)
+          .mul(bank.maintLiabWeight.sub(usdcBank.maintAssetWeight))
```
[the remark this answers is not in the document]
microwavedcola1: done

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,469 @@
...
+          .div(usdcBank.price);
```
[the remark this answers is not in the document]
microwavedcola1: done

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,469 @@
...
+        // Abs liab/borrow
+        const maxTokenHealthLiab = a.account
```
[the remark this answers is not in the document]
microwavedcola1: done

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,469 @@
...
+          bank.mint.toBase58(),
+        )
+      : { priceImpactPct: 0, outAmount: 0 };
+
+    // Sum of all assets which would be acquired in exchange for also acquiring
+    // liabs by the liqor, who would immediately want to reduce to 0
+    // Assuming assets need to be sold to USDC
+    const assets = mangoAccountsWithHealth.reduce((sum, a) => {
+      if (bank.maintAssetWeight.eq(ZERO_I80F48())) {
+        return sum.add(ZERO_I80F48());
```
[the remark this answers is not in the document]
microwavedcola1: correct, I guess in that case I should just assume that all of this asset (with asset-weight 0) can/will be taken by iqor

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,469 @@
...
+      }
+      // How much would health increase for every unit liab moved to liqor
+      const tokenAssetHealthContrib = ONE_I80F48()
+        .sub(bank.liquidationFee)
+        .mul(bank.price)
+        .mul(bank.maintAssetWeight);
```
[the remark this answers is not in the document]
microwavedcola1: thanks, yeah, I see the issue now, 

> we should just loop over all liabs and gradually fake-liquidate them.

can I simplify this to just choose the initLiabWeight of the token in the group with the smallest initLiabWeight? This way max asset would be liquidated

ts/client/scripts/risk.ts
microwavedcola1: done

ts/client/scripts/risk.ts
```diff
@@ -132,13 +144,14 @@ async function computePriceImpactForLiqor(
...
         // How much would health increase for every unit liab moved to liqor
-        const tokenLiabHealthContrib = ONE_I80F48()
-          .sub(bank.liquidationFee)
-          .mul(bank.price)
-          .mul(bank.maintLiabWeight.sub(usdcBank.maintAssetWeight))
-          .div(usdcBank.price);
+        // liabprice * (liabweight - (1+fee)*assetweight)
+        const tokenLiabHealthContrib = bank.price.mul(
+          bank.initLiabWeight.sub(
+            ONE_I80F48().add(bank.liquidationFee).mul(usdcBank.initAssetWeight),
```
ckamm, line 150: Minor: What if the liab itself is USDC? (a lot of our borrows are) Then the asset weight of the related token would be worse than the USDC one.

But using assetweight 1 is an ok upper bound.

ts/client/scripts/risk.ts
```diff
@@ -185,18 +196,27 @@ async function computePriceImpactForLiqor(
...
-        .mul(bank.maintAssetWeight);
+        .mul(bank.initAssetWeight);
+      bank.price.mul(
+        Array.from(group.banksMapByTokenIndex.values())
+          .flat()
+          .map((bank) => bank.initLiabWeight)
+          .reduce((prev, curr) => (prev.lt(curr) ? prev : curr))
+          .div(ONE_I80F48().add(bank.liquidationFee))
+          .sub(bank.initAssetWeight),
+      );
```
ckamm, line 212: I don't understand this bit. Does it have a side effect? Shouldn't tokenAssetHealthContrib be different? (according to the comment you added?)

ts/client/scripts/risk.ts
```diff
@@ -185,18 +196,27 @@ async function computePriceImpactForLiqor(
...
       // Abs collateral/asset
       const maxTokenHealthAsset = a.account
         .getTokenBalance(bank)
         .max(ZERO_I80F48());
+      if (bank.initAssetWeight.eq(ZERO_I80F48())) {
+        return sum.add(maxTokenHealthAsset);
```
ckamm, line 218: why would all asset be added? shouldn't that still depend on the health and how much health is gained by liquidating the assetweight=0 token to offset a borrow?

ckamm (commented):
Some large files `dev` and `mc-ckamm` are committed.

ts/client/scripts/risk.ts
microwavedcola1: wow ok, I had written the new version on a new line, and forgot to remove the old one, done

ts/client/scripts/risk.ts
microwavedcola1: correct, with the change in equation actually this condition has become redundant, removed

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,574 @@
...
+  // High level solana defi stats
+  const response = await fetch('https://cache.jup.ag/stats/day');
+  const res = await response.json();
+
+  const USDC_MINT = 'EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v';
+  const usdcBank = group.getFirstBankByMint(new PublicKey(USDC_MINT));
+
+  // For each token
+  for (const banks of Array.from(group.banksMapByMint.values())) {
+    // TODO USDC
```
ckamm, line 127: minor: what's TODO exactly?

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,574 @@
...
+    const bank = banks[0];
+
+    const onChainPrice = (
+      await (
+        await fetch(`https://price.jup.ag/v4/price?ids=${bank.mint}`)
+      ).json()
+    )['data'][bank.mint.toBase58()]['price'];
+
+    if (bank.reduceOnly > 0 || bank.tokenIndex === usdcBank.tokenIndex) {
+      continue;
```
ckamm, line 137: minor: could do it above onChainPrice fetching

Is it really right to skip reduceOnly tokens? Like DUAL for instance could still be liquidated...

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,574 @@
...
+        // How much would health increase for every unit liab moved to liqor
+        // liabprice * (liabweight - (1+fee)*assetweight)
+        const tokenLiabHealthContrib = bank.price.mul(
+          bank.initLiabWeight.sub(
+            ONE_I80F48().add(bank.liquidationFee).mul(usdcBank.initAssetWeight),
+          ),
+        );
+        // Abs liab/borrow
+        const maxTokenLiab = a.account
+          .getTokenBalance(bank)
```
ckamm, line 155: This is going to cause problems when accounts have funds on the account and on the serum oo account. 

Larger accounts often have token liabs that are offset by free funds in a oo account. I think we should sum up the actual balance and the serum free amounts. All uses of getTokenBalance() need to be evaluated.

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,574 @@
...
+    // liabs by the liqor, who would immediately want to reduce to 0
+    // Assuming assets need to be sold to USDC
+    const assets = mangoAccountsWithHealth.reduce((sum, a) => {
+      // How much would health increase for every unit liab moved to liqor
+      // assetprice * (liabweight/(1+liabliqfee) - assetweight)
+      const tokenAssetHealthContrib = bank.price.mul(
+        Array.from(group.banksMapByTokenIndex.values())
+          .flat()
+          .map((bank) => bank.initLiabWeight)
+          .reduce((prev, curr) => (prev.lt(curr) ? prev : curr))
```
ckamm, line 205: This would be easier for me to read if the min initLiabWeight was computed separately beforehand.

ts/client/scripts/risk.ts
```diff
@@ -0,0 +1,574 @@
...
+          .div(ONE_I80F48().add(bank.liquidationFee))
+          .sub(bank.initAssetWeight),
+      );
+      // Abs collateral/asset
+      const maxTokenHealthAsset = a.account
+        .getTokenBalance(bank)
```
ckamm, line 211: token balance

ts/client/scripts/risk.ts
microwavedcola1: was unintended, removing

ts/client/scripts/risk.ts
microwavedcola1: see change in mango account ts

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -231,9 +231,24 @@ export class MangoAccount {
...
    * @returns native balance for a token, is signed
    */
-  public getTokenBalance(bank: Bank): I80F48 {
+  public getTokenBalance(group: Group, bank: Bank): I80F48 {
     const tp = this.getToken(bank.tokenIndex);
-    return tp ? tp.balance(bank) : ZERO_I80F48();
+    if (tp) {
+      const bal = tp.balance(bank);
+      for (const serum3Market of Array.from(
+        group.serum3MarketsMapByMarketIndex.values(),
```
ckamm, line 239: my understanding is that `values()` gives you and iterator, so making an Array out of it just for iterating should be redundant

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -231,9 +231,24 @@ export class MangoAccount {
...
+      )) {
+        const oo = this.serum3OosMapByMarketIndex.get(serum3Market.marketIndex);
```
ckamm, line 241: mh. if we need the group only to get all the serum3 market indexes, why don't we just iterate over the key, value of serum3OosMapByMarketIndex?

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -231,9 +231,24 @@ export class MangoAccount {
    * @param bank
    * @returns native balance for a token, is signed
    */
-  public getTokenBalance(bank: Bank): I80F48 {
+  public getTokenBalance(group: Group, bank: Bank): I80F48 {
```
ckamm, line 234: I think having this kind of function is needed, but it should probably be separate from getTokenBalance? Existing users (such as the force-close) could break by changing these semantics.

ts/client/src/accounts/mangoAccount.ts
microwavedcola1: we also need to know the base token index and the quote token index

ts/client/src/accounts/mangoAccount.ts
microwavedcola1: done, I now have a new function called getEffectiveTokenBalance, and I have reset'ted the old getTokenBalance, I think in most cases the UI wouldn't always use getTokenBalance, getEffectiveTokenBalance is what I use now for health, risk, and various getMax* function families

ckamm (approved):

PR #568, opened 2023-05-03
Clarkeni/oracle confidence

nicholasgodfreyclarke:
Add oracle confidence to update funding logs (with a new V2 anchor event).
Unfortunately the meaning of the confidence value is different for different oracle providers so I also log the oracle type so they can be interpreted appropriately on the other end.
Also the confidence data types are u64, f64 and I80F48 - so I cast them all to I80F48.
Please let me know if you would like changes.
I believe there is another deployment this week - would like to target that if we can.

### DIFF
--- a/programs/mango-v4/src/state/perp_market.rs
+++ b/programs/mango-v4/src/state/perp_market.rs
@@ -11,7 +11,7 @@ use crate::logs::PerpUpdateFundingLogV2;
 use crate::state::orderbook::Side;
 use crate::state::{oracle, TokenIndex};
 
-use super::{orderbook, OracleConfig, Orderbook, StablePriceModel, DAY_I80F48};
+use super::{orderbook, OracleConfig, OracleState, Orderbook, StablePriceModel, DAY_I80F48};
 
 pub type PerpMarketIndex = u16;
 
@@ -246,7 +246,7 @@ impl PerpMarket {
         staleness_slot: Option<u64>,
     ) -> Result<I80F48> {
         require_keys_eq!(self.oracle, *oracle_acc.key());
-        let (price, _, _, _) = oracle::oracle_price_and_meta(
+        let (price, _) = oracle::oracle_price_and_state(
             oracle_acc,
             &self.oracle_config,
             self.base_decimals,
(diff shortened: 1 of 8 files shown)


programs/mango-v4/src/state/perp_market.rs
```diff
@@ -256,13 +256,13 @@ impl PerpMarket {
         Ok(price)
     }
 
-    pub fn oracle_price_and_slot(
+    pub fn oracle_price_and_meta(
         &self,
         oracle_acc: &impl KeyedAccountReader,
         staleness_slot: Option<u64>,
-    ) -> Result<(I80F48, u64)> {
+    ) -> Result<(I80F48, u64, I80F48, oracle::OracleType)> {
```
ckamm, line 263: suggestion: make a struct OracleState (or something) with price, slot, confidence, type so this function can have a clearer return va lue

programs/mango-v4/src/state/orderbook/book.rs
```diff
@@ -138,12 +138,13 @@ impl<'a> Orderbook<'a> {
                 matched_order_changes.push((best_opposing.handle, new_best_opposing_quantity));
             }
 
+            let seq_num = event_queue.header.seq_num;
```
ckamm, line 141: this change is baffling to me, because my version of book.rs already has it applied. I don't see why github shows it again

ckamm:
Your PR adds some things to the IDL that disappear when I rerun update-local-idl after the merge (duplications of "CheckLiquidatable"). It might be worthwhile to rebase on top op dev, drop the first commit of this PR (since it was merged already) and generate the idl from there.

programs/mango-v4/src/state/perp_market.rs
nicholasgodfreyclarke: Thanks for that suggestion - I have implemented it.

programs/mango-v4/src/state/orderbook/book.rs
nicholasgodfreyclarke: User error on my part - now resolved.

nicholasgodfreyclarke:
I had incorrectly created my branch while not in the dev branch. I was able to resolve with the help of your suggestion to rebase.
I also implemented your suggestion to return an OracleState struct.

ckamm (approved):

PR #574, opened 2023-05-05
ts port (Configurable perp settle token #550)

microwavedcola1:


ts/client/src/accounts/healthCache.ts
```diff
@@ -160,168 +203,185 @@ export class HealthCache {
...
+  }
 
+  healthAssetsAndLiabsStableLiabs(healthType: HealthType): {
+    assets: I80F48;
+    liabs: I80F48;
+  } {
+    return this.assetsAndLiabs(healthType, false);
+  }
+
+  public assetsAndLiabs(
```
ckamm, line 301: this is called healthAssetsAndLiabs upstream

ts/client/src/accounts/healthCache.ts
```diff
@@ -528,30 +586,9 @@ export class HealthCache {
...
-    console.log(
-      ` assets ${this.assets(HealthType.init)}, liabs ${this.liabs(
-        HealthType.init,
-      )}, `,
-    );
-    console.log(` health(HealthType.init) ${this.health(HealthType.init)}`);
-    console.log(
-      ` healthRatio(HealthType.init) ${this.healthRatio(HealthType.init)}`,
-    );
+    // TODO
```
ckamm, line 591: todo!

ts/client/src/accounts/healthCache.ts
```diff
@@ -1332,11 +1396,9 @@ export class Serum3Info {
     const baseFree = I80F48.fromI64(oo.baseTokenFree);
     // NOTE: referrerRebatesAccrued is not declared on oo class, but the layout
     // is aware of it
-    const quoteFree = I80F48.fromI64(
-      oo.quoteTokenFree.add((oo as any).referrerRebatesAccrued),
-    );
```
ckamm, line 1337: important change!

ckamm (commented):
I didn't review in detail, but skimmed..

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -375,20 +378,22 @@ export class MangoAccount {
...
   /**
    * Sum of all negative assets.
+   * TODO: healthType = HealthType.maint breaks UI assumption
    * @returns liabs, in native quote
    */
-  public getLiabsValue(group: Group, healthType?: HealthType): I80F48 {
+  public getLiabsValue(group: Group, healthType = HealthType.maint): I80F48 {
     const hc = HealthCache.fromMangoAccount(group, this);
-    return hc.liabs(healthType);
+    return hc.assetsAndLiabs(healthType, false).liabs;
```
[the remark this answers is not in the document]
microwavedcola1: I have now made the HealthCache.ts again accept undefined health type like it previously used to. Though I am not sure if I should use stable assets or stable liabs when UI wants to know assets value, and when it wants to know liabs value

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -378,22 +378,20 @@ export class MangoAccount {
...
   /**
    * Sum of all negative assets.
-   * TODO: healthType = HealthType.maint breaks UI assumption
    * @returns liabs, in native quote
    */
-  public getLiabsValue(group: Group, healthType = HealthType.maint): I80F48 {
+  public getLiabsValue(group: Group): I80F48 {
     const hc = HealthCache.fromMangoAccount(group, this);
-    return hc.assetsAndLiabs(healthType, false).liabs;
+    return hc.healthAssetsAndLiabs(undefined, false).liabs;
```
ckamm, line 394: I think getAssetsValue and getLiabsValue need to use a consistent version of healthAssetsAndLiabs(), otherwise comparing the two doesn't make sense (and the ui will show them next to each other)

ckamm: Ah, if all weights are set to 1, the both versions are the same!

microwavedcola1: I sitll made them consistent

PR #581, opened 2023-05-10
document unused field

microwavedcola1:


ts/client/src/clientIxParamBuilder.ts
```diff
@@ -157,7 +157,6 @@ export interface IxGateParams {
   Serum3CloseOpenOrders: boolean;
   Serum3CreateOpenOrders: boolean;
   Serum3DeregisterMarket: boolean;
-  Serum3EditMarket: boolean;
```
microwavedcola1, line 160: note: This will break governance ui temporarily

ckamm:
Is this really necessary? What about just leaving a comment on the program code and removing the gating on the instruction? We could leave the ts-client side symbols (and even the logging in the instruction0 as they were.

microwavedcola1:
> Is this really necessary? What about just leaving a comment on the program code and removing the gating on the instruction? We could leave the ts-client side symbols (and even the logging in the instruction0 as they were.

its already unused

ckamm (approved):

PR #593, opened 2023-05-24
wip: liquidation price for perp positions

microwavedcola1:


ts/client/src/accounts/perp.ts
```diff
@@ -440,6 +441,12 @@ export class PerpMarket {
     return parseFloat(price.toString()) * this.priceLotsToUiConverter;
   }
 
+  public priceLotsToNative(price: BN): I80F48 {
+    return I80F48.fromI64(price)
+      .mul(I80F48.fromI64(this.quoteLotSize))
+      .div(I80F48.fromI64(this.baseLotSize));
```
mschneider, line 447: division should be in BN not in fixedp to avoid rounding issues

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+    client: MangoClient,
+    group: Group,
+    perpMarket: PerpMarket,
+    mangoAccount: MangoAccount,
+    forceReload = false,
+  ): Promise<I80F48> {
+    const pp = mangoAccount.getPerpPosition(perpMarket.perpMarketIndex);
+    if (!pp) {
+      throw new Error(
+        `Cannot compute liquidation price for a position which doesn't, perpMarketIndex - ${perpMarket.perpMarketIndex}`,
```
ckamm, line 1690: "doesn't exist"

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+    sortedOo: PerpOrder[],
+    tillPriceExclusive: I80F48,
+  ): I80F48 {
+    const settleTokenBank = group.getFirstBankByTokenIndex(
+      perpMarket.settleTokenIndex,
+    );
+    const settleTokenPrice = settleTokenBank.price;
+
+    // For constants B, T, C and final liqudation price equation, refer to
+    // https://www.notion.so/mango-markets/Mango-v4-Perp-Liq-Price-b7dde5e70c724d07934a21dc0aa79518?pvs=4
```
ckamm, line 1774: imo this should either be public, or the explanation should be inlined

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+              I80F48.fromI64(perpMarket.baseLotSize),
+            ),
+          ),
+        );
+        qpn = qpn.sub(ret.quoteNative);
+      }
+    }
+
+    if (wbpn.isZero()) {
+      return MINUS_ONE_I80F48();
```
ckamm, line 1823: If it always returns -1, even for shorts, then shouldn't that mean that prevLp will be -1 and it won't consider what happens when the asks > oracle price execute?

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+          tillPriceExclusive,
+        );
+        wbpn = perpMarket.maintBaseAssetWeight.mul(
+          baseNative.add(
+            I80F48.fromI64(ret.sizeLots).mul(
+              I80F48.fromI64(perpMarket.baseLotSize),
+            ),
+          ),
+        );
+        qpn = qpn.sub(ret.quoteNative);
```
ckamm, line 1818: it sounds like this `if` could be simplified if the `ret` value had sizeLots and quoteNative with appropriate sign, so they could always be added in.

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+              I80F48.fromI64(perpMarket.baseLotSize),
+            ),
+          ),
+        );
+        qpn = qpn.add(ret.quoteNative);
+      } else {
+        const ret = PerpPosition.getAggregatePerpOrderSizeAndQuoteNative(
+          perpMarket,
+          sortedOo,
+          PerpOrderSide.ask,
```
ckamm, line 1808: this `ask` is suspicious

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+    )!;
+    ppToBeErased.marketIndex =
+      PerpPosition.PerpMarketIndexUnset as PerpMarketIndex;
+    const healthCacheWoPerpPosition = HealthCache.fromMangoAccount(
+      group,
+      mAClone,
+    );
+
+    // Unused, TODO what to do with this?
+    // const basePrice = qpn.neg().div(wbpn);
```
ckamm, line 1839: I don't get this TODO

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+
+    const healthContributionWithoutSettleToken =
+      healthCacheWoPerpPosition.healthExcludingToken(
+        HealthType.maint,
+        perpMarket.settleTokenIndex,
+      ); // C
```
ckamm, line 1845: For this, you don't need the WoPerpPosition, since you exclude the whole token index anyway. If you drop the right perp market in the filter() in line 1858, you might not need to manually wipe the perp pos at all.

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+        healthCacheWoPerpPosition.getOrCreateTokenInfoIndex(settleTokenBank)
+      ].balanceSpot;
+    const otherPerpsSettledInSettleTokenUhuPnl =
+      healthCacheWoPerpPosition.perpInfos
+        .filter((pi) => pi.settleTokenIndex == perpMarket.settleTokenIndex)
+        .reduce(
+          (sum, pi) => sum.add(pi.healthUnsettledPnl(HealthType.maint)),
+          ZERO_I80F48(),
+        );
+    const openBookFree = mangoAccount.getSerum3TokenFreeBalance(
```
ckamm, line 1863: balanceSpot already includes this, no need to add it again

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+
+  public static getAggregatePerpOrderSizeAndQuoteNative(
+    perpMarket: PerpMarket,
+    orders: PerpOrder[],
+    side: PerpOrderSide,
+    tillPriceExclusive: I80F48,
+  ): { sizeLots: BN; quoteNative: I80F48 } {
+    const filtered = orders.filter((oo) =>
+      side == PerpOrderSide.bid
+        ? perpMarket.priceLotsToNative(oo.priceLots).gte(tillPriceExclusive)
```
ckamm, line 1744: you say exclusive, but this looks pretty inclusive?

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+        : perpMarket.priceLotsToNative(oo.priceLots).lte(tillPriceExclusive),
+    );
+    return {
+      sizeLots: filtered.reduce((sum, a) => sum.add(a.sizeLots), new BN(0)),
+      quoteNative: filtered.reduce(
+        (sum, a) =>
+          sum.add(
+            I80F48.fromI64(a.sizeLots)
+              .mul(I80F48.fromI64(perpMarket.baseLotSize))
+              .mul(perpMarket.priceLotsToNative(a.priceLots)),
```
ckamm, line 1754: minor, alternatively: sizeLots * priceLots * quoteLots

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+          ZERO_I80F48(),
+        );
+    const openBookFree = mangoAccount.getSerum3TokenFreeBalance(
+      group,
+      settleTokenBank,
+    );
+    const settleTokenContribution = settleTokenSpot
+      .add(otherPerpsSettledInSettleTokenUhuPnl)
+      .add(openBookFree)
+      .add(qpn); // T
```
ckamm, line 1870: It might be preferable to build C and T by subtraction, instead of constructively.

```
C = maintHealth 
    - tokenContribution(effectiveTokenPos[settleToken])
T = tokenContribution(effectiveTokenPos[settleToken]) 
    - perpPos.hupnl 
    + quote pos native
```

That way this code doesn't need to know as much about how health is constructed, just about perp health.

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1628,6 +1676,243 @@ export class PerpPosition {
...
+      );
+
+      if (prevLp && pp.getBasePositionNative(perpMarket).isPos()) {
+        if (!nextLp.gt(prevLp)) {
+          break;
+        }
+      } else if (prevLp && !nextLp.lt(prevLp)) {
+        break;
+      }
+      prevLp = nextLp;
```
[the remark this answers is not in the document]
microwavedcola1: prevLp is null at start

ts/client/src/accounts/healthCache.ts
```diff
@@ -270,6 +297,18 @@ export class HealthCache {
...
+  public healthExcludingToken(
+    healthType: HealthType,
+    skipTokenIndex: TokenIndex,
+  ): I80F48 {
+    const tokenBalances = this.effectiveTokenBalancesInternal(
+      healthType,
+      false,
+      skipTokenIndex,
+    );
+    return this.healthSum(healthType, tokenBalances);
```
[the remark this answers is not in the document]
microwavedcola1: hmm do you mean that this skipToken thing is not enough? Because even if skip this token, it was probably involved in spot markets, which we still considered

image

PR #596, opened 2023-05-25
referrerRebatesAccrued - these accrue to the ui now

microwavedcola1:


ts/client/src/accounts/mangoAccount.ts
```diff
@@ -368,11 +368,7 @@ export class MangoAccount {
       // is aware of it
```
ckamm, line 368: the note about referrerRebatesAccrued can be deleted

ckamm (approved):

PR #599, opened 2023-06-04
Add loan amount to WithdrawLoanLog. Add withdraw loan logging to toke…

nicholasgodfreyclarke:
…n_force_close_borrows_with_token

### DIFF
--- a/programs/mango-v4/src/logs.rs
+++ b/programs/mango-v4/src/logs.rs
@@ -248,19 +248,6 @@ pub enum LoanOriginationFeeInstruction {
     TokenWithdraw,
 }
 
-#[derive(PartialEq, Copy, Clone, Debug, AnchorSerialize, AnchorDeserialize)]
-#[repr(u8)]
-pub enum LoanOriginationFeeInstructionV2 {
-    Unknown,
-    LiqTokenBankruptcy,
-    LiqTokenWithToken,
-    Serum3LiqForceCancelOrders,
-    Serum3PlaceOrder,
-    Serum3SettleFunds,
-    TokenWithdraw,
-    TokenForceCloseBorrowsWithToken,
-}
-
 #[event]
 pub struct WithdrawLoanOriginationFeeLog {
     pub mango_group: Pubkey,
@@ -277,7 +264,7 @@ pub struct WithdrawLoanLog {
     pub token_index: u16,
     pub loan_amount: i128,
     pub loan_origination_fee: i128,
-    pub instruction: LoanOriginationFeeInstructionV2,
+    pub instruction: LoanOriginationFeeInstruction,
 }
 
 #[event]
--- a/programs/mango-v4/src/logs.rs
+++ b/programs/mango-v4/src/logs.rs
@@ -248,6 +248,19 @@ pub enum LoanOriginationFeeInstruction {
     TokenWithdraw,
 }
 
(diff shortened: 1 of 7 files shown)


nicholasgodfreyclarke:
For the upcoming marketing push we would like to see amount of loan on withdrawals.
I also added this logging to token_force_close_borrows_with_token (it was missed previously).
I created a new enum for the instruction source (LoanOriginationFeeInstructionV2). Similar to how we keep events immutable (and prefer to create new events) - I think the same applies to enums referenced in events.

programs/mango-v4/src/state/bank.rs
```diff
@@ -473,7 +481,7 @@ impl Bank {
         with_loan_origination_fee: bool,
         allow_dusting: bool,
         now_ts: u64,
-    ) -> Result<(bool, I80F48)> {
+    ) -> Result<WithdrawResult> {
```
ckamm, line 484: This is nice

programs/mango-v4/src/state/bank.rs
```diff
@@ -486,13 +494,21 @@ impl Bank {
                     self.dust += new_native_position;
                     self.indexed_deposits -= position.indexed_position;
                     position.indexed_position = I80F48::ZERO;
-                    return Ok((false, I80F48::ZERO));
+                    return Ok(WithdrawResult {
+                        position_is_active: true,
```
ckamm, line 498: `false`

programs/mango-v4/src/state/bank.rs
```diff
@@ -528,19 +548,25 @@ impl Bank {
...
+        Ok(WithdrawResult {
+            position_is_active,
             loan_origination_fee,
-            false,
-            !position.is_in_use(),
-            now_ts,
-        )?;
-
-        Ok((position_is_active, loan_origination_fee))
+            loan_amount: already_borrowed_native_amount,
```
ckamm, line 568: Hmm, this is a tricky case. The borrow has already happened before and withdraw has returned a WithdrawResult with loan_origination_fee == 0 and loan_amount > 0. To be consistent, we should return loan_amount = 0 here.

Otherwise there'd be double-counting of loans!

programs/mango-v4/src/state/bank.rs
```diff
@@ -563,9 +589,13 @@ impl Bank {
         position: &mut TokenPosition,
         native_amount: I80F48,
         now_ts: u64,
-    ) -> Result<(bool, I80F48)> {
+    ) -> Result<WithdrawResult> {
```
ckamm, line 592: Mh, ok

programs/mango-v4/src/logs.rs
```diff
@@ -248,6 +248,19 @@ pub enum LoanOriginationFeeInstruction {
     TokenWithdraw,
 }
 
+#[derive(PartialEq, Copy, Clone, Debug, AnchorSerialize, AnchorDeserialize)]
+#[repr(u8)]
+pub enum LoanOriginationFeeInstructionV2 {
```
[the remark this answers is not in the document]
nicholasgodfreyclarke: Thanks - that's good to know.

programs/mango-v4/src/state/bank.rs
nicholasgodfreyclarke: Sorry!

ckamm (approved):

PR #601, opened 2023-06-05
Expose instruction builders for Openbook

mschneider:
* create OO
* place order
* cancel all

also removed unneeded result/error return type that was technically dead code

### DIFF
--- a/bin/keeper/src/taker.rs
+++ b/bin/keeper/src/taker.rs
@@ -138,7 +138,7 @@ pub async fn loop_blocking_price_update(
         log::info!("{} Updated price is {:?}", token_name, fresh_price.price);
         if let Ok(mut price) = price.write() {
             *price = I80F48::from_num(fresh_price.price)
-                / I80F48::from_num(10u64.pow(-fresh_price.expo as u32));
+                / I80F48::from_num(10f64.powi(-fresh_price.expo));
         }
     }
 }
@@ -157,6 +157,29 @@ pub async fn loop_blocking_orders(
         .unwrap();
     log::info!("Cancelled orders - {:?} for {}", orders, market_name);
 
+    let market_index = mango_client
+        .context
+        .serum3_market_indexes_by_name
+        .get(&market_name)
+        .unwrap()
+        .clone();
+
+    let s3 = mango_client
+        .context
+        .serum3_markets
+        .get(&market_index)
+        .unwrap()
+        .clone();
+
+    let base_decimals = mango_client
+        .context
+        .token(s3.market.base_token_index)
+        .decimals as i32;
+    let quote_decimals = mango_client
+        .context
+        .token(s3.market.quote_token_index)
+        .decimals as i32;
+
     loop {
         interval.tick().await;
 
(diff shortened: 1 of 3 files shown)


lib/client/src/client.rs
```diff
@@ -429,36 +429,66 @@ impl MangoClient {
...
+        let quote_info = self.context.token(serum3_info.market.quote_token_index);
+        let base_info = self.context.token(serum3_info.market.base_token_index);
+
+        Serum3Data {
+            market_index,
+            market: serum3_info,
+            quote: quote_info,
+            base: base_info,
+        }
+    }
```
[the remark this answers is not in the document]
mschneider: yeah moved them over, looks a little bit better now

bin/keeper/src/taker.rs
```diff
@@ -157,32 +157,42 @@ pub async fn loop_blocking_orders(
...
-                    anyhow::bail!("Price RwLock PoisonError!");
-                }
-            };
+            let fresh_price = price.read().unwrap().to_num::<f64>();
+            let bid_price = fresh_price * 1.1;
 
-            let fresh_price = fresh_price.to_num::<f64>();
+            let bid_price_lots =
+                bid_price * 10f64.powi(quote_decimals - base_decimals) * s3.coin_lot_size as f64
+                    / s3.pc_lot_size as f64;
```
ckamm, line 187: might be nice to have price_ui_to_serum_lots() or something (I'm unclear on whether `fresh_price` is the ui price or native price or something else)

lib/client/src/client.rs
```diff
@@ -515,15 +481,21 @@ impl MangoClient {
         account: &MangoAccountValue,
         market_index: Serum3MarketIndex,
         side: Serum3Side,
-        price: f64,
-        size: f64,
+        limit_price: u64,
+        max_base_qty: u64,
+        max_native_quote_qty_including_fees: u64,
```
ckamm, line 486: There definitely should be a simple ix constructor like this one; but wondering if we should keep mc's idea of having a simplified one with ui numbers. 

I guess this client is for advanced users who're ok figuring out "quote_qty_incl_fees" themselves. 

I'd like if it was `limit_price_lots` and `max_base_qty_lots` to be clearer about the units of these two.

lib/client/src/client.rs
```diff
@@ -532,65 +504,9 @@ impl MangoClient {
...
-            (price * ((10u64.pow(s3.quote.decimals as u32) * s3.market.coin_lot_size) as f64))
-                as u64
-                / (10u64.pow(s3.base.decimals as u32) * s3.market.pc_lot_size)
-        };
-        // https://github.com/project-serum/serum-ts/blob/master/packages/serum/src/market.ts#L1333
-        let max_base_qty =
-            { (size * 10u64.pow(s3.base.decimals as u32) as f64) as u64 / s3.market.coin_lot_size };
-        let max_native_quote_qty_including_fees = {
-            fn get_fee_tier(msrm_balance: u64, srm_balance: u64) -> u64 {
-                if msrm_balance >= 1 {
```
ckamm, line 535: oh yeah, this was very very out of date

bin/keeper/src/taker.rs
mschneider: openbook v2 sdk is going to have that imo

lib/client/src/client.rs
mschneider: ui numbers should be another layer on top, ideally a rest api

PR #603, opened 2023-06-11
Add price to WithdrawLoanLog

nicholasgodfreyclarke:
For performance reasons it is desirable to have prices logged in the WithdrawLoanLog.
I am using the same event (not incrementing it) as I hope this can be merged before the previous WithdrawLoanLog event is deployed.


ckamm (commented):
Let's talk about deprecating Serum3SettleFunds in favor of V2

ckamm (approved):

PR #610, opened 2023-06-15
add disclaimer to bots

microwavedcola1:


ckamm (approved):

bin/liquidator/README.md
```diff
@@ -1,3 +1,29 @@
+## Disclaimer
+
+The following open source code contains an automated bot designed to perform liquidations. Please note that the use of this code is at your own risk and responsibility.
```
mschneider, line 3: would rather call it an example to document how to use the API

microwavedcola1: done

PR #612, opened 2023-06-16
deploy->dev

microwavedcola1:
port changes done for ts client back to dev


ckamm:
I think we should revert my node-fetch + Procfile patch, because we're fixed it properly in dev already. Agree?

PR #615, opened 2023-06-16
Fix ts client margin trade method to also work for delegates

microwavedcola1:


ts/client/src/client.ts
```diff
@@ -2843,6 +2843,13 @@ export class MangoClient {
     userDefinedAlts: AddressLookupTableAccount[];
     flashLoanType: FlashLoanType;
   }): Promise<TransactionSignature> {
+    const isDelegate = (
+      this.program.provider as AnchorProvider
+    ).wallet.publicKey.equals(mangoAccount.delegate);
```
ckamm, line 2848: minor: isDelegate() might be a useful helper function

ckamm (approved):

PR #618, opened 2023-06-19
mc-ckamm/stop loss liq - ts client

microwavedcola1:


ts/client/src/client.ts
```diff
@@ -655,28 +663,6 @@ export class MangoClient {
 
   // MangoAccount
 
-  public async getOrCreateMangoAccount(
```
microwavedcola1, line 658: removed this finally, confirmed not used in ui

ts/client/src/client.ts
```diff
@@ -705,44 +691,43 @@ export class MangoClient {
     return await this.sendAndConfirmTransactionForGroup(group, [ix]);
   }
 
-  public async createAndFetchMangoAccount(
```
microwavedcola1, line 708: also removed this, confirmed not used in the ui

ts/client/src/client.ts
```diff
@@ -812,21 +797,90 @@ export class MangoClient {
   }
 
   public async getMangoAccount(
-    mangoAccount: MangoAccount | PublicKey,
```
microwavedcola1, line 815: simplified this

ts/client/scripts/archive/devnet-user-2.ts
```diff
@@ -71,9 +72,14 @@ async function main() {
 
   // create + fetch account
   console.log(`Creating mangoaccount...`);
-  const mangoAccount = await client.getOrCreateMangoAccount(group);
+  const mangoAccount = (await client.getMangoAccountForOwner(
```
ckamm, line 75: this is no longer creating accounts, right? but the whole script is in archive/ so...

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1677,6 +1688,76 @@ export class PerpOoDto {
...
+      dto.priceLimit,
+      dto.pricePremiumBps,
+      dto.takerFeeBps,
+      dto.makerFeeBps,
+      dto.buyTokenIndex as TokenIndex,
+      dto.sellTokenIndex as TokenIndex,
+      dto.isActive == 1,
+      dto.priceThresholdType == 0
+        ? 'priceOverThreshold'
+        : 'priceUnderThreshold',
```
ckamm, line 1709: Shouldn't this use `TokenConditionalSwapPriceThresholdType`?

ts/client/src/client.ts
ckamm: This might break existing users, but I agree being able to call this on a MangoAccount was weird

ts/client/src/client.ts
```diff
@@ -812,21 +797,105 @@ export class MangoClient {
...
+  ): MangoAccount {
+    const decodedMangoAccount = this.program.coder.accounts.decode(
+      'mangoAccount',
+      ai.data,
+    );
+
+    // Start reading tokenConditionalSwaps from this point in account buffer
+    const start =
+      // anchor headers
+      8 +
```
ckamm, line 833: I don't like this, and was going to suggest using `coder.size()`, but when I checked that code, it looked like it doesn't actually work with accounts that contain Vecs (assumes their size is 1 byte!).

Maybe re-encode the account and use that size?

How about we make a MangoAccountV2 type that you could try decoding the data as?

ts/client/src/client.ts
```diff
@@ -872,23 +937,43 @@ export class MangoClient {
     ownerPk: PublicKey,
     loadSerum3Oo = false,
   ): Promise<MangoAccount[]> {
+    const anchroHeaderFilter: {
```
ckamm, line 940: anchro

ts/client/src/client.ts
```diff
@@ -872,23 +937,43 @@ export class MangoClient {
...
-          },
-        },
-        {
-          memcmp: {
-            bytes: ownerPk.toBase58(),
-            offset: 40,
-          },
+          filters: [
+            {
+              memcmp: {
```
ckamm, line 954: suspect you could just do `memcmp: anchorHeaderFilter`

ts/client/src/client.ts
```diff
@@ -872,23 +937,43 @@ export class MangoClient {
     ownerPk: PublicKey,
     loadSerum3Oo = false,
   ): Promise<MangoAccount[]> {
+    const anchroHeaderFilter: {
+      offset: number;
+      bytes: string;
+    } = this.program.account.mangoAccount.coder.accounts.memcmp(
+      this.program.account.mangoAccount['_idlAccount'].name,
```
ckamm, line 944: I don't get the advantage over 'mangoAccount'

ts/client/src/client.ts
```diff
@@ -906,22 +991,27 @@ export class MangoClient {
...
-            offset: 8,
-          },
-        },
-        {
-          memcmp: {
-            bytes: delegate.toBase58(),
-            offset: 104,
-          },
+          filters: [
+            {
```
ckamm, line 998: is missing the discriminator filter

ts/client/src/client.ts
```diff
@@ -938,16 +1028,21 @@ export class MangoClient {
...
+      await this.program.provider.connection.getProgramAccounts(
+        this.programId,
         {
-          memcmp: {
-            bytes: group.publicKey.toBase58(),
-            offset: 8,
-          },
+          filters: [
+            {
+              memcmp: {
```
ckamm, line 1036: discr filter

ckamm:
merged into `ckamm/stop-loss-liq` manually, comments still relevant

PR #624, opened 2023-06-26
Use FillEventLog3 for closed pnl logging

Lou-Kamades:


ckamm (commented):
This doesn't log only closed pnl:
- What if the taker had 0 base lots before. They wouldn't produce realized pnl at all.
- Generally if their number of base lots increases away from 0, there's no realized pnl.
- If the fill makes the position change sign, there's partial realized pnl and partially a new position.

We already do realized pnl tracking in update_trade_stats(), see realized_trade_pnl_native. Possibly comparing this value before and after the fill would give you what you want?

programs/mango-v4/src/logs.rs
```diff
@@ -137,6 +137,33 @@ pub struct FillLogV2 {
...
+    pub maker_timestamp: u64,
+
+    pub taker: Pubkey,
+    pub taker_client_order_id: u64,
+    pub taker_fee: f32,
+
+    pub price: i64,
+    pub quantity: i64, // number of base lots
+    pub maker_closed_pnl: f64,
+    pub taker_closed_pnl: f64,
```
ckamm, line 164: minor: maybe a comment saying that this is in settle-token-native units.

ckamm (approved):

PR #625, opened 2023-06-26
Mc/perp liq price 2

microwavedcola1:


ts/client/src/accounts/healthCache.ts
```diff
@@ -1159,6 +1160,52 @@ export class HealthCache {
...
+    const hc = HealthCache.fromMangoAccount(group, mangoAccount);
+    const perpMarket = group.getPerpMarketByMarketIndex(
+      perpPosition.marketIndex,
+    );
+
+    function healthAfterPriceChange(newPrice: I80F48): I80F48 {
+      const gClone = cloneDeep(group);
+      gClone.getPerpMarketByMarketIndex(perpMarket.perpMarketIndex)._price =
+        newPrice;
+      const hc = HealthCache.fromMangoAccount(gClone, mangoAccount);
```
ckamm, line 1178: we could save a bunch of compute here by not rebuilding the HealthCache from scratch and just overriding the perp price. But what you're doing now is true black-box use of HealthCache, so maybe most stable

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1432,6 +1438,27 @@ export class PerpPosition {
     );
   }
 
+  public getLiquidationPrice(group: Group, mangoAccount: MangoAccount): I80F48 {
```
ckamm, line 1441: maybe return I80F48 | null or something, so there doesn't need to be a I80F48 marker value for "none"?

ckamm (approved):

PR #628, opened 2023-06-30
Openbook v2 integration

godmodegalactus:
Initial commites,

adding dependency, creating initial instruction accounts, updating entry point to do nothing.


programs/mango-v4/src/accounts_ix/openbook_v2_cancel_all_orders.rs
```diff
@@ -0,0 +1,47 @@
...
+        has_one = openbook_v2_program,
+        has_one = openbook_v2_market_external,
+    )]
+    pub openbook_v2_market: AccountLoader<'info, OpenbookV2Market>,
+
+    pub openbook_v2_program: Program<'info, OpenbookV2>,
+    #[account(mut)]
+    pub openbook_v2_market_external: AccountLoader<'info, Market>,
+
+    #[account(mut)]
```
ckamm, line 39: Add a note saying that events/bids/asks accounts are just forwarded to a ob_v2 cpi call, and are validated there.

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_all_orders.rs
```diff
@@ -0,0 +1,47 @@
...
+    #[account(
+        has_one = group,
+        constraint = account.load()?.is_operational() @ MangoError::AccountIsFrozen
+        // owner is checked at #1
+    )]
+    pub account: AccountLoader<'info, MangoAccountFixed>,
+    pub owner: Signer<'info>,
+
+    #[account(mut)]
+    pub open_orders: AccountLoader<'info, OpenOrdersAccount>,
```
ckamm, line 26: document how we know that this open_orders is the right one

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_order.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use crate::error::*;
+use crate::state::*;
+use openbook_v2::{
+    program::OpenbookV2,
+    state::{BookSide, Market, OpenOrdersAccount},
+};
+
+#[derive(Accounts)]
+pub struct OpenbookV2CancelOrder<'info> {
```
ckamm, line 11: Am I mistaken, or do Cancel and CancelAll look identical? Let's have only one accounts struct for both, imo. (should have done that for perp too)

programs/mango-v4/src/accounts_ix/openbook_v2_create_open_orders.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use crate::error::*;
+use crate::state::*;
+use openbook_v2::{
+    program::OpenbookV2,
+    state::{Market, OpenOrdersAccount},
+};
+
+#[derive(Accounts)]
+#[instruction(account_num: u32, open_orders_count: u8)]
```
ckamm, line 11: account_num may be confused with the account_num passed to account_create. Maybe open_orders_account_num? Looks like the open_orders_count doesn't need to be seen here.

programs/mango-v4/src/accounts_ix/openbook_v2_create_open_orders.rs
```diff
@@ -0,0 +1,54 @@
...
+        has_one = openbook_v2_market_external,
+    )]
+    pub openbook_v2_market: AccountLoader<'info, OpenbookV2Market>,
+
+    pub openbook_v2_program: Program<'info, OpenbookV2>,
+    pub openbook_v2_market_external: AccountLoader<'info, Market>,
+
+    // initialized by this instruction via cpi to openbook_v2
+    #[account(
+        init,
```
ckamm, line 39: is it really up to us to init this account? I suspect it'd be the openbookv2 program that inits it!

programs/mango-v4/src/accounts_ix/openbook_v2_liq_force_cancel_orders.rs
```diff
@@ -0,0 +1,69 @@
...
+    )]
+    pub openbook_v2_market: AccountLoader<'info, OpenbookV2Market>,
+
+    pub openbook_v2_program: Program<'info, OpenbookV2>,
+
+    #[account(mut)]
+    pub openbook_v2_market_external: AccountLoader<'info, Market>,
+
+    #[account(mut)]
+    pub market_bids: AccountLoader<'info, BookSide>,
```
ckamm, line 41: again, validation comments would be good. Also, I don't thing we really need to use `AccountLoader<'info, BookSide>` right? we never interact with the data there directly, so could just be an UncheckedAccount that we forward to the openbook program instruction?

programs/mango-v4/src/accounts_ix/openbook_v2_place_taker_order.rs
```diff
@@ -0,0 +1,73 @@
...
+use crate::state::*;
+use anchor_lang::prelude::*;
+use anchor_spl::token::{Token, TokenAccount};
+use openbook_v2::{
+    program::OpenbookV2,
+    state::{BookSide, Market},
+};
+
+#[derive(Accounts)]
+pub struct OpenbookV2PlaceTakerOrder<'info> {
```
ckamm, line 11: possibly PlaceOrder and PlaceTakerOrder have the same accounts object

programs/mango-v4/src/accounts_ix/openbook_v2_settle_funds.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    // token_index and bank.vault == vault is validated inline at #3
+    #[account(mut, has_one = group)]
+    pub quote_bank: AccountLoader<'info, Bank>,
+    #[account(mut)]
+    pub quote_vault: Box<Account<'info, TokenAccount>>,
+    #[account(mut, has_one = group)]
+    pub base_bank: AccountLoader<'info, Bank>,
+    #[account(mut)]
+    pub base_vault: Box<Account<'info, TokenAccount>>,
```
ckamm, line 60: also pass the quote and base oracle, like in Serum3SettleFundsV2

programs/mango-v4/src/lib.rs
```diff
@@ -1125,6 +1125,93 @@ pub mod mango_v4 {
...
+        force_close_opt: Option<bool>,
+    ) -> Result<()> {
+        Ok(())
+    }
+
+    pub fn openbook_v2_deregister_market(ctx: Context<OpenbookV2DeregisterMarket>) -> Result<()> {
+        Ok(())
+    }
+
+    pub fn openbook_v2_create_open_orders(ctx: Context<OpenbookV2CreateOpenOrders>) -> Result<()> {
```
ckamm, line 1152: create open orders must take at least `account_num: u32, open_orders_count: u8` to match the accounts struct

ckamm: (and I don't yet understand what open_orders_count is for)

programs/mango-v4/src/lib.rs
```diff
@@ -1125,6 +1125,93 @@ pub mod mango_v4 {
...
+    }
+
+    pub fn openbook_v2_cancel_all_orders(
+        ctx: Context<OpenbookV2CancelAllOrders>,
+        limit: u8,
+    ) -> Result<()> {
+        Ok(())
+    }
+
+    pub fn openbook_v2_settle_funds(ctx: Context<OpenbookV2SettleFunds>) -> Result<()> {
```
ckamm, line 1204: possibly there should be a `fees_to_dao: bool` arg, like in serum settle funds v2

programs/mango-v4/src/state/group.rs
```diff
@@ -190,6 +194,17 @@ pub enum IxGate {
     TokenForceCloseBorrowsWithToken = 49,
     PerpForceClosePosition = 50,
     GroupWithdrawInsuranceFund = 51,
+    OpenbookV2CancelAllOrders = 52,
```
ckamm, line 197: Will need to update these numbers when rebasing

programs/mango-v4/src/state/openbook_v2_market.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    pub market_index: OpenbookV2MarketIndex,
+
+    pub bump: u8,
+
+    pub padding2: [u8; 5],
+
+    pub registration_time: u64,
+
+    pub reserved: [u8; 128],
```
ckamm, line 33: recommend bumping this to at least 512 reserved bytes, we'll only have few of these

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_all_orders.rs
godmodegalactus: Will add a check on market.

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_all_orders.rs
```diff
@@ -0,0 +1,47 @@
...
+        has_one = openbook_v2_market_external,
+    )]
+    pub openbook_v2_market: AccountLoader<'info, OpenbookV2Market>,
+
+    pub openbook_v2_program: Program<'info, OpenbookV2>,
+    #[account(mut)]
+    pub openbook_v2_market_external: AccountLoader<'info, Market>,
+
+    #[account(mut)]
+    pub market_bids: AccountLoader<'info, BookSide>,
```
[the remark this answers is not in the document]
godmodegalactus: Yeah could be an issue later, May be we can typedef it of ObV2BookSide

programs/mango-v4/src/accounts_ix/openbook_v2_create_open_orders.rs
godmodegalactus: Yes i have moved it to unchecked account. we cannot init openbook_v2 programs account with anchor like this.

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_all_orders.rs
godmodegalactus: I added some constraints by loading the open-book market and check if it corresponds to asks, bids and event queue.

programs/mango-v4/src/accounts_ix/openbook_v2_liq_force_cancel_orders.rs
godmodegalactus: adding some constaints.

programs/mango-v4/src/lib.rs
godmodegalactus: Yes not needed `open_orders_count`.

programs/mango-v4/src/lib.rs
godmodegalactus: Done.

programs/mango-v4/src/state/group.rs
godmodegalactus: Done

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_order.rs
```diff
@@ -0,0 +1,55 @@
...
+        constraint = account.load()?.is_operational() @ MangoError::AccountIsFrozen
+        // owner is checked at #1
+    )]
+    pub account: AccountLoader<'info, MangoAccountFixed>,
+
+    pub authority: Signer<'info>,
+
+    #[account(
+        mut,
+        constraint = open_orders.load()?.market == openbook_v2_market_external.key()
```
ckamm, line 29: All accounts structs: I don't think this is the actual constraint we need. It would still be possible to pass in any openorders account, even for another users. Instead of this constraint, there should be a check that verifies that this openorders account is listed on the mangoaccount.

On the serum3 account this typically is done in instruction code and these get a comment like
```
/// CHECK: Validated inline by checking against the pubkey stored in the account at #2
```
I feel that would be appropriate here as well.

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_order.rs
```diff
@@ -0,0 +1,55 @@
...
+    #[account(
+        constraint = group.load()?.is_ix_enabled(IxGate::OpenbookV2CancelOrder) @ MangoError::IxIsDisabled,
+    )]
+    pub group: AccountLoader<'info, Group>,
+
+    #[account(
+        mut,
+        has_one = group,
+        constraint = account.load()?.is_operational() @ MangoError::AccountIsFrozen
+        // owner is checked at #1
```
ckamm, line 21: All accounts structs: `// authority is checked at #1`

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_order.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    #[account(
+        has_one = bids,
+        has_one = asks,
+    )]
+    pub openbook_v2_market_external: AccountLoader<'info, Market>,
+
+    #[account(mut)]
+    /// CHECK: bids will be checked by openbook_v2
+    pub bids: AccountLoader<'info, ObV2BookSize>,
```
ckamm, line 50: `ObV2BookSide`

The rename doesn't confuse the idl? I'm still unclear on whether we'll ever need to deserialize these.

programs/mango-v4/src/accounts_ix/openbook_v2_edit_market.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use crate::state::*;
+
+#[derive(Accounts)]
+#[instruction(market_index: OpenbookV2MarketIndex)]
+pub struct OpenbookV2EditMarket<'info> {
+    #[account(
+        // group <-> admin relation is checked at #1
+        constraint = group.load()?.openbook_v2_supported(),
+        constraint = group.load()?.admin == admin.key(),
```
ckamm, line 11: the comment doesn't make sense, since the admin pubkey is already checked here

programs/mango-v4/src/accounts_ix/openbook_v2_liq_force_cancel_orders.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub openbook_v2_market: AccountLoader<'info, OpenbookV2Market>,
+
+    pub openbook_v2_program: Program<'info, OpenbookV2>,
+
+    #[account(
+        has_one = bids,
+        has_one = asks,
+        has_one = event_queue,
+        constraint = openbook_v2_market_external.load()?.base_vault == market_base_vault.key(),
+        constraint = openbook_v2_market_external.load()?.quote_vault == market_quote_vault.key(),
```
ckamm, line 45: these are typically better as `address = openbook_v2_market_external.load()?.base_vault` style constraints on the target accounts. However, I'm not yet sure it's good to propagate all these constraints to the mango interface. We don't do that for serum.

The main pros are, imo, that it helps the anchor clients to autodeduce more accounts, and we might give better error messages to users. However, as a drawback, a bunch of openbook v2 specifics leak into mango and all these checks are done twice (once in the mango program and once when cpi-ing into obv2). So it's technically wasteful.

programs/mango-v4/src/accounts_ix/openbook_v2_place_taker_order.rs
```diff
@@ -0,0 +1,87 @@
+use crate::error::*;
+use crate::state::*;
+use anchor_lang::prelude::*;
+use anchor_spl::token::{Token, TokenAccount};
+use openbook_v2::{program::OpenbookV2, state::Market};
+
+#[derive(Accounts)]
+pub struct OpenbookV2PlaceTakerOrder<'info> {
+    #[account(
+        constraint = group.load()?.is_ix_enabled(IxGate::OpenbookV2PlaceOrder) @ MangoError::IxIsDisabled,
```
ckamm, line 10: Should this have a PlaceTakerOrder enum?

programs/mango-v4/src/accounts_ix/openbook_v2_settle_funds.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub openbook_v2_market: AccountLoader<'info, OpenbookV2Market>,
+
+    pub openbook_v2_program: Program<'info, OpenbookV2>,
+
+    #[account(
+        mut,
+        constraint = openbook_v2_market_external.load()?.base_vault == market_base_vault.key(),
+        constraint = openbook_v2_market_external.load()?.quote_vault == market_quote_vault.key(),
+        constraint = openbook_v2_market_external.load()?.base_mint == base_bank.load()?.mint,
+        constraint = openbook_v2_market_external.load()?.quote_mint == quote_bank.load()?.mint,
```
ckamm, line 45: the last two checks are correct, but also unnecessary imo: if the mints don't match, the instruction would definitely fail.

Instead, we should check that the banks are for the expected token indexes stored in the openbook_v2_market (like for serum3, there it's done in the instruction code)

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_order.rs
godmodegalactus: This constraint just checks if the openorders belongs to the same market right. So I guess passing openorders for another user should work.

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_order.rs
godmodegalactus: replaced all remaining owners.

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_order.rs
godmodegalactus: Ok I will simply make them UncheckedAccounts, and then we will see later if we need to deserialize them.

programs/mango-v4/src/accounts_ix/openbook_v2_liq_force_cancel_orders.rs
godmodegalactus: I added these constraints so that we can we do not do wasteful computation before cpi into open book. But yeah ideally these kind of errors will be filtered out in testing phase. I can remove these constaints.

programs/mango-v4/src/accounts_ix/openbook_v2_edit_market.rs
godmodegalactus: Removed the comment.

programs/mango-v4/src/accounts_ix/openbook_v2_cancel_order.rs
```diff
@@ -18,17 +15,15 @@ pub struct OpenbookV2CancelOrder<'info> {
...
 
     pub authority: Signer<'info>,
 
-    #[account(
-        mut,
-        constraint = open_orders.load()?.market == openbook_v2_market_external.key()
-    )]
-    pub open_orders: AccountLoader<'info, OpenOrdersAccount>,
+    #[account(mut)]
+    /// CHECK: openorders will be checked by openbook_v2
```
ckamm: Sorry, need to comment on this again!

Yes, the openbook v2 cpi will verify that this openorders account is for the given openbook market.

However, the key point is that mango will need to verify that this open_orders account is owned by the mango account passed above. That's essential - otherwise a user could pass in another user's openorders account and do something with it.

programs/mango-v4/src/accounts_ix/openbook_v2_register_market.rs
```diff
@@ -19,11 +19,7 @@ pub struct OpenbookV2RegisterMarket<'info> {
     /// CHECK: Can register a market for any openbook_v2 program
     pub openbook_v2_program: Program<'info, OpenbookV2>,
 
-    /// CHECK: Can register any openbook_v2 market
-    #[account(
-        constraint = openbook_v2_market_external.load()?.base_mint == base_bank.load()?.mint,
-        constraint = openbook_v2_market_external.load()?.quote_mint == quote_bank.load()?.mint,
-    )]
```
ckamm: for register_market we do need to check the mints like this, can you bring it back? (sorry!)

ckamm (approved):

PR #637, opened 2023-07-09
Leverage & Balance Sheet helpers

mschneider:


programs/mango-v4/src/health/client.rs
```diff
@@ -1458,4 +1520,105 @@ mod tests {
             assert!((assets.to_num::<f64>() - 2.0 * (10.0 * 1.2 + 2.0 * 0.8)) < 0.01);
         }
     }
+
+    #[test]
+    fn test_leverage() {
+        let buffer = MangoAccount::default_for_tests().try_to_vec().unwrap();
+        let mut account = MangoAccountValue::from_bytes(&buffer).unwrap();
+        account.ensure_token_position(0).unwrap();
+        account.ensure_token_position(1).unwrap();
```
ckamm, line 1529: `account` is unused

programs/mango-v4/src/health/client.rs
```diff
@@ -446,6 +446,58 @@ impl HealthCache {
...
+
+        for serum_info in self.serum3_infos.iter() {
+            let quote = &self.token_infos[serum_info.quote_info_index];
+            let base = &self.token_infos[serum_info.base_info_index];
+            assets += serum_info.reserved_base * base.prices.oracle;
+            assets += serum_info.reserved_quote * quote.prices.oracle;
+        }
+
+        for perp_info in self.perp_infos.iter() {
+            // TODO: multiply with quote price
```
ckamm, line 472: yeah, multiply both the quote and base contributions with the price for perp_info.settle_token_index

programs/mango-v4/src/health/client.rs
mschneider: right

programs/mango-v4/src/health/client.rs
mschneider: why base? base oracle should be base/USD no?

ckamm: It's the way perp margining and perp settlement works right now. The oracle is base/USD, but the settlement is in USDC. If you have 1 base long and the price goes up by 1 USD then you can settle 1 USDC. Hence the USDC price needs to be applied to the base oracle here.

We talked about switching from "perps on SOL/USD with USDC payouts" to "perps on SOL/USDC perps with USDC payouts" because the latter should be more interesting to users when the USDC price is unstable. But that has not yet been done.

ckamm (approved):

PR #640, opened 2023-07-11
move math for health_cache.equity from client.rs to cache.rs to make it available in on-chain programs

grooviegermanikus:


mschneider:
please merge directly to dev, after https://github.com/blockworks-foundation/mango-v4/pull/637

PR #643, opened 2023-07-12
Mc/perp close all

microwavedcola1:


ts/client/src/client.ts
```diff
@@ -2265,6 +2265,52 @@ export class MangoClient {
     return await this.sendAndConfirmTransactionForGroup(group, [ix]);
   }
 
+  public async perpCloseAll(
+    group: Group,
+    mangoAccount: MangoAccount,
+    slippage = 0.01, // 1%, 100bps
+  ): Promise<TransactionSignature> {
+    if (mangoAccount.perpActive().length > 0) {
+      throw new Error(`No perp positions found.`);
```
ckamm, line 2274: You mean == 0

ts/client/src/client.ts
```diff
@@ -2265,6 +2265,52 @@ export class MangoClient {
...
+
+    const ixs = await Promise.all(
+      mangoAccount.perpActive().map(async (pa) => {
+        const pm = group.getPerpMarketByMarketIndex(pa.marketIndex);
+
+        return await this.perpPlaceOrderV2Ix(
+          group,
+          mangoAccount,
+          pa.marketIndex,
+          pa.basePositionLots.gt(new BN(0))
```
ckamm, line 2292: minor: compute           `pa.basePositionLots.gt(new BN(0))` before the call

ts/client/src/client.ts
```diff
@@ -2265,6 +2265,52 @@ export class MangoClient {
...
+            ? PerpOrderSide.ask
+            : PerpOrderSide.bid,
+          pm.uiPrice *
+            (pa.basePositionLots.gt(new BN(0)) ? 1 + slippage : 1 - slippage), // Try to cross the spread to guarantee matching
```
ckamm, line 2296: isn't this inverted? If we're long, we should be willing to sell at slippage _less_ than oracle price?

ts/client/src/client.ts
```diff
@@ -2265,6 +2265,52 @@ export class MangoClient {
...
+  public async perpCloseAll(
+    group: Group,
+    mangoAccount: MangoAccount,
+    slippage = 0.01, // 1%, 100bps
+  ): Promise<TransactionSignature> {
+    if (mangoAccount.perpActive().length > 0) {
+      throw new Error(`No perp positions found.`);
+    }
+
+    if (mangoAccount.perpActive().length > 8) {
```
ckamm, line 2277: optionally save a lot of compute by using a health region (at least leave this as a suggestion here)

ts/client/src/client.ts
```diff
@@ -2642,6 +2688,56 @@ export class MangoClient {
...
+    for (const pa of mangoAccount.perpActive()) {
+      const pm = group.getPerpMarketByMarketIndex(pa.marketIndex);
+      const candidates = pm.getSettlePnlCandidates(
+        client,
+        group,
+        allMangoAccounts,
+        pa.getEquityUi(pm) > 0 ? 'negative' : 'positive',
+        2,
+      );
+      if ((await candidates).length == 0) {
```
ckamm, line 2714: I'd prefer if the await was in the `const candidates = await ...`

ts/client/src/client.ts
```diff
@@ -2642,6 +2688,56 @@ export class MangoClient {
...
+    // This is optimistic, since we might find the same opponent candidate for all markets,
+    // and they have might not be able to settle at some point due to safety limits
+    // Future: correct way to do is, to apply the settlement on a copy and then move to next position
+    for (const pa of mangoAccount.perpActive()) {
+      const pm = group.getPerpMarketByMarketIndex(pa.marketIndex);
+      const candidates = pm.getSettlePnlCandidates(
+        client,
+        group,
+        allMangoAccounts,
+        pa.getEquityUi(pm) > 0 ? 'negative' : 'positive',
```
ckamm, line 2711: is `getEquityUi` the right function for checking if there's settleable pnl?

Suggest giving a name to pa.getEquityUi(pm) > 0, since it pops up often

What if there's no settleable pnl (unlikely, I know)

microwavedcola1: Whis is what the ui did which I just blindly copy pasted, oof, replaced with 
image

ckamm: up here it still says getEquity

PR #644, opened 2023-07-13
Fees to dao

microwavedcola1:


programs/mango-v4/src/instructions/dao_withdraw_fees_token.rs
```diff
@@ -0,0 +1,21 @@
...
+
+pub fn dao_withdraw_fees_token(ctx: Context<DaoWithdrawFeesToken>) -> Result<()> {
+    let group = ctx.accounts.group.load()?;
+    let mut bank = ctx.accounts.bank.load_mut()?;
+
+    let group_seeds = group_seeds!(group);
+    let amount = bank.collected_fees_native.floor().to_num::<u64>();
+    token::transfer(
+        ctx.accounts.transfer_ctx().with_signer(&[group_seeds]),
+        bank.collected_fees_native.floor().to_num::<u64>(),
```
ckamm, line 15: use `amount`

programs/mango-v4/src/instructions/dao_withdraw_fees_token.rs
```diff
@@ -0,0 +1,21 @@
...
+    )?;
+
+    bank.collected_fees_native -= amount_i80f48;
```
ckamm, line 18: amount_i80f48 does not exist

ckamm (commented):
Two bigger-picture issues:

- If we decrease collected_fees and fees_settled, then the charts for mango markets lifetime fees will stop working. Possibly there should be new fields that store how many tokens were transfered out? (but would need some detail comments, because it's weird that fees_accrued moves to fees_settled but does not move to fees_taken_out etc)
- It's possible that the bank vault does not have enough tokens. It should probably just take as many tokens as are available.
- With multi-bank, it would be helpful if we could take collected_fees for usdc_bank1 out of usdc_bank2's token vault.

microwavedcola1:
> Two bigger-picture issues:
> 
> * If we decrease collected_fees and fees_settled, then the charts for mango markets lifetime fees will stop working. Possibly there should be new fields that store how many tokens were transfered out? (but would need some detail comments, because it's weird that fees_accrued moves to fees_settled but does not move to fees_taken_out etc)
> * It's possible that the bank vault does not have enough tokens. It should probably just take as many tokens as are available.
> * With multi-bank, it would be helpful if we could take collected_fees for usdc_bank1 out of usdc_bank2's token vault.

pt 2 done

ckamm (approved):

PR #651, opened 2023-07-27
Client: use createWithSeed when wrapping SOL

vovacodes:
Using `SystemProgram.createAccountWithSeed` instead of `SystemProgram.createAccount` eliminates the need in an ephemeral signer for the the wrapping SOL transaction. 

This makes the transaction compatible with Smart Wallets that use delayed transaction execution mechanism, e.g. Fuse Wallet

For more in-depth explanation see section "Ephemeral Signers" here: https://fusedocs.framer.website/docs/fuse-intergration-for-app-devs

### DIFF
--- a/ts/client/src/client.ts
+++ b/ts/client/src/client.ts
@@ -1248,33 +1248,41 @@ export class MangoClient {
     let wrappedSolAccount: Keypair | undefined;
     let preInstructions: TransactionInstruction[] = [];
     let postInstructions: TransactionInstruction[] = [];
-    const additionalSigners: Signer[] = [];
     if (mintPk.equals(NATIVE_MINT)) {
-      wrappedSolAccount = new Keypair();
+      // Generate a random seed for wrappedSolAccount.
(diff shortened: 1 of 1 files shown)


ts/client/src/client.ts
```diff
@@ -1248,33 +1248,41 @@ export class MangoClient {
     let wrappedSolAccount: Keypair | undefined;
     let preInstructions: TransactionInstruction[] = [];
     let postInstructions: TransactionInstruction[] = [];
-    const additionalSigners: Signer[] = [];
     if (mintPk.equals(NATIVE_MINT)) {
-      wrappedSolAccount = new Keypair();
+      // Generate a random seed for wrappedSolAccount.
+      const seed = Keypair.generate().publicKey.toBase58().slice(0, 32);
+      // Calculate a publicKey that will be controlled by the `mangoAccount.owner`.
+      const wrappedSolAccount = await PublicKey.createWithSeed(
```
vovacodes, line 1255: This is where the bug happens, I'm creating a new `const` rather than assigning to the `let` defined above, on line 1248 🤦

PR #652, opened 2023-08-02
use indexed position in token conditional balance logs

Lou-Kamades:
previously we were logging native amount, which does not match other `TokenBalanceLog`s


nicholasgodfreyclarke:
We've temporarily disabled logging of token balances associated with token conditional swap trigger until this fix can be deployed (to avoid large spikes in performance metrics).

microwavedcola1 (approved):

PR #653, opened 2023-08-03
token conditional swaps, fixes from review

microwavedcola1:


ts/client/src/client.ts
```diff
@@ -3461,66 +3625,7 @@ export class MangoClient {
...
-  //       ),
-  //     );
-  //   }
-  //   ixs.push(tcsIx);
-
-  //   return await this.sendAndConfirmTransactionForGroup(group, ixs);
-  // }
-
-  public async tokenConditionalSwapCreate(
+  public async tokenConditionalSwapCreateOld(
```
ckamm, line 3628: `Old` postfix is a bit suspicious. Maybe `Raw` because it matches what the program does without helpful logic?

ts/client/src/client.ts
```diff
@@ -3374,64 +3372,230 @@ export class MangoClient {
...
+      upperLimit = thresholdPrice;
+    }
+
     const expiryTimestampBn =
       expiryTimestamp !== null ? new BN(expiryTimestamp) : U64_MAX_BN;
 
+    pricePremium = group.getPriceImpactByTokenIndex(
+      sellBank.tokenIndex,
+      maxSellUi * sellBank.uiPrice,
+    );
```
ckamm, line 3586: unconditionally overrides the passed-in pricePremium; also, isn't the price impact dependent on both, the sell _and_ buy bank? It looks like this would go quite wrong if you want to buy BONK for USDC and then USDC's price impact gets used.

ts/client/src/client.ts
```diff
@@ -3374,64 +3372,230 @@ export class MangoClient {
...
+    const pricePremiumFraction = pricePremium > 0 ? pricePremium : 0.03;
```
ckamm, line 3587: I'm still suspicious of that 3% default. And also I suspect you want pricePremium to be in percent, so it should be divided by 100?

ts/client/src/client.ts
```diff
@@ -3374,64 +3372,230 @@ export class MangoClient {
...
+      tcsIntention == 'TakeProfitOnDeposit' ||
+      tcsIntention == 'StopLossOnBorrow' ||
+      (tcsIntention == null && thresholdPrice > sellTokenPerBuyTokenPrice)
+    ) {
+      lowerLimit = thresholdPrice;
+      upperLimit = Number.MAX_SAFE_INTEGER;
+    } else {
+      lowerLimit = 0;
+      upperLimit = thresholdPrice;
+    }
```
ckamm, line 3578: minor: I was surprised to see this here instead of in the wrapper functions? They know what they want and could pass us the appropriate upper/lower limit?

ts/client/src/client.ts
```diff
@@ -3374,64 +3372,230 @@ export class MangoClient {
...
-    pricePremiumFraction: number | null,
+    sellBank: Bank,
+    buyBank: Bank,
+    thresholdPriceUi: number,
+    thresholdPriceInSellPerBuyToken: boolean,
+    maxBuyUi: number | null,
+    maxSellUi: number | null,
+    pricePremium: number | null,
+    allowCreatingDeposits: boolean | null,
+    allowCreatingBorrows: boolean | null,
```
ckamm, line 3386: When you want to take profit on a deposit, you don't need to pass in maxBuy -- you want to buy as much as you can!

Similarly, why can you override allowCreatingBorrows? You're taking profit, you shouldn't accidentally create a borrow in the token!

(same arguments for the other functions)

ts/client/src/client.ts
```diff
@@ -3374,64 +3372,230 @@ export class MangoClient {
...
     expiryTimestamp: number | null,
   ): Promise<TransactionSignature> {
-    const buyBank: Bank = group.getFirstBankByMint(buyMintPk ?? USDC_MINT);
-    const sellBank: Bank = group.getFirstBankByMint(sellMintPk);
+    if (maxSellUi == null && account.getTokenBalanceUi(sellBank) < 0) {
```
ckamm, line 3389: why check only when maxSellUi is null?

ts/client/src/client.ts
microwavedcola1: changed to

```

    if (!pricePremium) {
      pricePremium = Math.max(
        group.getPriceImpactByTokenIndex(
          sellBank.tokenIndex,
          maxSellUi * sellBank.uiPrice,
        ),
        group.getPriceImpactByTokenIndex(
          buyBank.tokenIndex,
          maxSellUi * sellBank.uiPrice,
        ),
      );
    }
```

I am thinking of just hardcoding to 10k$ instead of using `maxSellUi * sellBank.uiPrice`, the liquidator would just swap in chunks, or?

ts/client/src/client.ts
microwavedcola1: pricePremiumFraction is passed on as is to program from here, would I still need to divide by 100? if my intention was 3%...

ts/client/src/client.ts
microwavedcola1: because in this function, normally, we wouldnt care about maxBuyUi, we are doing a take profit, so we would want to this on entire deposit

ts/client/src/client.ts
microwavedcola1: I think like 75% of below code will be repeated, hence I did it here

```
if (!thresholdPriceInSellPerBuyToken) {
      thresholdPriceUi = 1 / thresholdPriceUi;
    }

    let lowerLimit, upperLimit;
    const thresholdPrice = toNativeSellPerBuyTokenPrice(
      thresholdPriceUi,
      sellBank,
      buyBank,
    );
    const sellTokenPerBuyTokenPrice = buyBank.price
      .div(sellBank.price)
      .toNumber();

    if (
      tcsIntention == 'TakeProfitOnDeposit' ||
      tcsIntention == 'StopLossOnBorrow' ||
      (tcsIntention == null && thresholdPrice > sellTokenPerBuyTokenPrice)
    ) {
      lowerLimit = thresholdPrice;
      upperLimit = Number.MAX_SAFE_INTEGER;
    } else {
      lowerLimit = 0;
      upperLimit = thresholdPrice;
    }
```

ts/client/src/client.ts
microwavedcola1: Fixed all of these, please check, I have only kept 

allowCreatingBorrows for tcsStopLossOnBorrow and tcsTakeProfitOnBorrow, in case you are using margin

ts/client/src/client.ts
ckamm: I meant dividing pricePremium, the 0.03 is fine.

ts/client/src/client.ts
ckamm: The price impact is relative to USDC right? Then I'd expect both impacts to be combined (get an approximation of the the full impact by swapping to USDC and then to the target). 

If they return a value in percent, then I'd expect `((1 + sell_impact/100) * (1 + buy_impact/100) - 1) * 100` to give the combined number.

Also the amounts should not both be `maxSellUi * sellBank.uiPrice` -- and actually maxSellUi might be inf. Need to get the min of maxBuy and maxSell (converted to the same token)

ts/client/src/client.ts
ckamm: If we take profit/stop loss a deposit, I think we should always check that the sell bank has a positive deposit - no matter if maxSell is specified or not.

ts/client/src/client.ts
ckamm: Looks good. Maybe rename allowCreatingBorrows to allowMargin?

ts/client/src/client.ts
microwavedcola1: Please see latest change 

```
if (!pricePremium) {
      const buyTokenPriceImpact = group.getPriceImpactByTokenIndex(
        buyBank.tokenIndex,
        5000,
      );
      const sellTokenPriceImpact = group.getPriceImpactByTokenIndex(
        sellBank.tokenIndex,
        5000,
      );
      pricePremium =
        ((1 + buyTokenPriceImpact / 100) * (1 + sellTokenPriceImpact / 100) -
          1) *
        100;
    }
    const pricePremiumFraction = pricePremium > 0 ? pricePremium / 100 : 0.03;
    ```

ckamm (approved):

PR #659, opened 2023-08-04
Tcs: Use fees config from Bank instead of Group

ckamm:
To give more flexibility to the DAO.


programs/mango-v4/src/instructions/token_register.rs
```diff
@@ -91,7 +91,9 @@ pub fn token_register(
         force_close: 0,
         padding: Default::default(),
         fees_withdrawn: 0,
-        reserved: [0; 2104],
+        token_conditional_swap_taker_fee_fraction: 0.0,
```
microwavedcola1, line 94: just to make sure I get this - so you went with 5 bps for trustless and 0 for trusted

microwavedcola1 (approved):

programs/mango-v4/src/instructions/token_register.rs
ckamm: My current idea is that "register" leaves most things null and expects a follow up with "edit", whereas "register_trustless" should come with full defaults.

Probably better if register also comes with defaults.

microwavedcola1 (approved):

PR #661, opened 2023-08-04
Client: allow PDA token account owners by default

vovacodes:
Disallowing PDA token owners currently causes errors when the user wallet is a Smart Wallet (PDA).

Until there's a good reason for having `allowOwnerOffCurve = false` by default, I suggest we change the defaults to `true`.


microwavedcola1:
Applied change manually to deploy branch

PR #662, opened 2023-08-07
Add estimatePrioritizationFee to the ts client

Lou-Kamades:


ts/client/src/client.ts
```diff
@@ -3694,4 +3711,55 @@ export class MangoClient {
...
+        return acc;
+      },
+      {},
+    );
+    const maximumFees = Object.values(maxFeeBySlot).sort(
+      (a, b) => a.slot - b.slot,
+    );
+
+    // take the EMA
+    const smoothingFactor = 2 / (maximumFees.length + 1);
```
ckamm, line 3755: Could you explain what's going on with this alpha? I'd expect it to be fixed because the discounting of older slot data would be fixed regardless of how many data points they are?

Lou-Kamades: Initially I figured we might want the weighting to be more aggressive if the amount of slots returned was low. For example imagine:
1. 20 slots with a large fee in slot #10
2. 150 slots with a large fee in slot #140

Assuming in both scenarios the slots are the most recent available, if we used the formula, we would estimate the fee in scenario 1 as quite a bit higher than scenario 2 (which might be reasonable since we have less information to base our estimate off of).

That said, the RPC always returns 150 slots of data so I think it makes sense to hardcode the value at 0.01. Since 2 / (150+1) = 0.0133.

Additionally, the pattern I'm seeing in the fee responses is that they are quite sparse, so even if less than 150 slots get returned for some reason, heavily discounting the past is a good strategy -- using the previous forumla would result in overestimation of fees

ckamm: Thinking about this again, an ema doesn't seem appropriate - in particular if the data is sparse. With your proposed params, if there are 150 datapoints the oldest datapoint would have a 13% weight, second oldest 0.2% and newest 1.3%.

It's genuinely hard to find a good value imo. The getRecentPrioritizationFees results just guarantee that there was a tx in the block with the specified fee - it might have been one that randomly paid an excessive fee.

I'm not convinced the EMA helps. How about just using the median over the most recent 20 slots or so? That'd solve outliers and only use fairly recent data.

Lou-Kamades: Sounds good, I settled on `Min(1, median)` -- afaik our keepers currently have a priority fee of 1 and it's quite effective since fees are so sparse

ts/client/src/client.ts
```diff
@@ -4193,4 +4213,50 @@ export class MangoClient {
...
+      .flat();
+    const uniqueWritableAccounts = uniq(
+      writableAccounts.map((x) => x.toBase58()),
+    )
+      .map((a) => new PublicKey(a))
+      .slice(0, MAX_RECENT_PRIORITY_FEE_ACCOUNTS);
+
+    const priorityFees = await this.connection.getRecentPrioritizationFees({
+      lockedWritableAccounts: uniqueWritableAccounts,
+    });
```
ckamm, line 4239: it might be possible for priorityFees to be empty if no recent blocks had transactions for the queried account, let's return 1?

ckamm (commented):
Last minor comment, then good

ckamm (approved):

PR #663, opened 2023-08-07
Delegate with expiry

microwavedcola1:


programs/mango-v4/src/instructions/account_edit.rs
```diff
@@ -18,28 +22,26 @@ pub fn account_edit(
...
     }
 
-    // unchanged -
-    // tokens
-    // serum3
-    // perps
-    // being_liquidated
-    // is_bankrupt
+    match (temporary_delegate_opt, temporary_delegate_expiry_opt) {
+        (Some(temporary_delegate), Some(temporary_delegate_expiry)) => {
```
ckamm, line 34: optional: This bundling of the two flags doesn't seem necessary. I'd be fine with allowing them to be set independently.

programs/mango-v4/src/state/mango_account.rs
```diff
@@ -257,9 +257,14 @@ pub struct MangoAccountFixed {
     pub buyback_fees_accrued_previous: u64,
     pub buyback_fees_expiry_timestamp: u64,
     pub next_token_conditional_swap_id: u64,
-    pub reserved: [u8; 200],
+    pub temporary_delegate: Pubkey,
+    pub temporary_delegate_expiry: u64,
+    pub reserved: [u8; 160],
```
ckamm, line 262: we're eating up bytes pretty quickly; luckily we could extend for more fixed storage if necessary at a later point

ckamm (approved):

PR #665, opened 2023-08-07
Pass all params to token_register

ckamm:


ts/client/src/clientIxParamBuilder.ts
```diff
@@ -2,6 +2,66 @@ import { BN } from '@coral-xyz/anchor';
...
+  maintAssetWeight: 0,
+  initAssetWeight: 0,
+  maintLiabWeight: 1.4,
+  initLiabWeight: 1.8,
+  liquidationFee: 0.2,
+  stablePriceDelayIntervalSeconds: 60 * 60,
+  stablePriceDelayGrowthLimit: 0.06,
+  stablePriceGrowthLimit: 0.0003,
+  minVaultToDepositsRatio: 0.2,
+  netBorrowLimitPerWindowQuote: 250_000_000_000,
```
microwavedcola1, line 55: Params below look very generous, and others look very conservative, should we revise?

```
  netBorrowLimitPerWindowQuote: 250_000_000_000,
  borrowWeightScaleStartQuote: 5_000_000_000,
  depositWeightScaleStartQuote: 5_000_000_000,
```

ckamm: I copied over the values from register_trustless. Could drop the new borrow limit per day to $5k by default in both?

microwavedcola1: yes, please change borrow limit to $5k

PR #670, opened 2023-08-09
liquidator: execute tcs with volume-weighted randomness

ckamm:


bin/liquidator/src/trigger_tcs.rs
```diff
@@ -57,15 +72,16 @@ fn tcs_has_plausible_premium(
...
+    context: &MangoGroupContext,
+    account_fetcher: &chain_data::AccountFetcher,
     tcs: &TokenConditionalSwap,
     token_swap_info: &token_swap_info::TokenSwapInfoUpdater,
     now_ts: u64,
 ) -> anyhow::Result<bool> {
-    Ok(!tcs.is_expired(now_ts)
-        && tcs_is_in_price_range(mango_client, tcs).await?
-        && tcs_has_plausible_premium(tcs, token_swap_info)?)
+    Ok(tcs.is_expired(now_ts)
```
microwavedcola1, line 82: ah, now you consider expired ones also, because they can be evicted and tcs slot can be made free from the mango account, or?

microwavedcola1 (approved):

bin/liquidator/src/trigger_tcs.rs
```diff
@@ -265,42 +291,119 @@ pub async fn maybe_execute_token_conditional_swap(
...
+    let buy_position = account
+        .token_position(tcs.buy_token_index)
+        .map(|p| p.native(&buy_bank))
+        .unwrap_or(I80F48::ZERO);
+    let sell_position = account
+        .token_position(tcs.sell_token_index)
+        .map(|p| p.native(&sell_bank))
+        .unwrap_or(I80F48::ZERO);
+
+    let base_price = buy_token_price / sell_token_price;
```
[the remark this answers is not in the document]
ckamm: It's not intuitive, but `(USDC/BUY) / (USDC/SELL) = SELL/BUY`

bin/liquidator/src/trigger_tcs.rs
ckamm: Yeah, the definition of "interesting" changed here. :P

I collect all tcs that are possible good to execute or expired, put their volume in the list and pick one. Then the execution function either closes or triggers.

PR #681, opened 2023-08-16
FlashLoan: Allow mango ix after flash loan end

ckamm:
This lifts an unnecessary restriction, making it possible to do things like flash loan and then liquidate with the resulting balances in a single transaction.

The "no mango ix after FlashLoanEnd" limitation was unnecessary: After the flash loan ends, the mango accounts are in a valid state again. And it's impossible to use a two FlashLoanEnd for a single FlashLoanBegin for the same reason that isolated FlashLoanEnds are rejected.

### DIFF
--- a/programs/mango-v4/src/instructions/flash_loan.rs
+++ b/programs/mango-v4/src/instructions/flash_loan.rs
@@ -146,14 +146,6 @@ pub fn flash_loan_begin<'key, 'accounts, 'remaining, 'info>(
 
             // Check that the mango program key is not used
             if ix.program_id == crate::id() {
-                // must be the last mango ix -- this could possibly be relaxed, but right now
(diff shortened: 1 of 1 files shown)


Lou-Kamades:
What's the reason for not allowing any other mango instructions in the middle of a flash loan? Naively, I'd think a major use case would be atomic liquidations like:
1. take out a flash loan
2. liquidate someone
3. swap away the 
4. repay the flash loan

ckamm:
@Lou-Kamades Mango instructions during a flash loan make it very hard to reason about the safety. Currently we can be assured that no banks or vaults or accounts change state during the flash loan, simpllifying the flow a lot.

We could relax that, but it'd need significant effort to make sure it's safe.

PR #691, opened 2023-08-25
wip: Add pre_fill_taker_base_position and

nicholasgodfreyclarke:
pre_fill_maker_base_position to fill log


nicholasgodfreyclarke:
Once the logs are deployed the scraper can compare the pre fill base position against the quantity filled and the direction to categorize trades into one of:
- open short
- open long
- increase short
- increase long
- close (partial) short
- close (partial) long

nicholasgodfreyclarke:
On second thought - may be better to just return an enum with the fill category directly from the rust program. Will change to this approach tomorrow.

nicholasgodfreyclarke:
Currently this PR defines the enum
```
pub enum FillCategory {
    OpenShort,
    OpenLong,
    IncreaseShort,
    IncreaseLong,
    DecreaseShort,
    DecreaseLong,
}
```
And then in the situation where base position is 100 and 200 is sold (final base position -100) this is defined as DecreaseLong (conceivably this could also be defined as OpenShort).

But alternatively you could also define more enum values
```
CloseShort
CloseLong
ShortToLong
LongToShort
```
Thoughts?

programs/mango-v4/src/instructions/perp_consume_events.rs
```diff
@@ -44,6 +44,41 @@ macro_rules! load_mango_account {
...
+    OpenLong,
+    IncreaseShort,
+    IncreaseLong,
+    DecreaseShort,
+    DecreaseLong,
+}
+
+fn get_fill_category(pre_fill_base_position: i64, side: Side) -> FillCategory {
+    // Note: if base position is 100 and 200 is sold (final base position -100)
+    // this is defined as DecreaseLong (conceivably this could also be defined as OpenShort)
```
ckamm, line 59: I'm highly sceptical of the usefulness of this classification.
- Separating out OpenLong/Short will make people try to filter for that, but like this comment suggests: it won't work when the user's position crosses 0. So you cannot rely on the open events.
- And because you can go from -100 to +100 in a DecreaseShort, you can't rely on Increase events either

So I expect this will lead to queries that don't really work.

How about just logging the post lots for maker and taker? Then you can easily query for fills that increase a long with `post_taker_lots > 0 && taker_side == bid` etc. It will make the queries slightly more complex, but they'd be more likely to do the right thing imo.

microwavedcola1: maybe we can split when it crosses into 2 events

nicholasgodfreyclarke: Thanks for the review.
Trying to get the best of both worlds here - I added some more enum states to cover all combinations of pre and post lots and bid/ask. Then also added in logging of the `post_taker_lots` and `post_maker_lots`.
My reasons is that the enum is helpful for convenience to someone querying the fills - but they can break it down further if they like by looking at the post lots. I can remove the enum though if you are not convinced.

programs/mango-v4/src/instructions/perp_consume_events.rs
```diff
@@ -44,6 +44,64 @@ macro_rules! load_mango_account {
...
+    OpenLong,
+    IncreaseShort,
+    IncreaseLong,
+    DecreaseShort,
+    DecreaseLong,
+    CloseShort,
+    CloseLong,
+    ShortToLong,
+    LongToShort,
+    NoChange,
```
ckamm, line 59: I'm still sceptical of the usefulness of this categorization.

If there was no ShortToLong and LongToShort I'd kind of see the point: "DecreaseShort" + quantity: 40 tells me a lot I want to know.

But since ShortToLong/LongToShort exist, it seems that most users of the categorization will will need to write code to split these up. For example, if you wanted to figure out closed pnl you'd need to write something like:
```
let long_decrease = match category {
    DecreaseLong | CloseLong => quantity,
    LongToShort => min(quantity, post_base + quantity),
    _ => 0,
};
```
versus the code if the categories didn't exist
```
let long_decrease = if side == Ask {
    max(0, min(quantity, post_base + quantity))
} else {
    0
};
```
Where the former doesn't seem easier to me.

Can you give me an example of a query you're thinking of where these categories would be helpful?

microwavedcola1: hmm so here is what I would actually use, sorry for letting this get dragged so much,

OpenLong - here I would be interested in knowing the fee paid, pnl would be 0
OpenShort - here I would be interested in knowing the fee paid, pnl would be 0
CloseLong - here I would be interested in knowing the fee paid, pnl would most probably be non 0
CloseShort - here I would be interested in knowing the fee paid, pnl would most probably be non 0

Basically I want to continuously track realized pnl including fees, and also if I am placing the correct maker orders and doing correct taker sides and how they are faring. Actually the strategy is very sensitive to pnl and fees

fwiw I am fine abandoning this, and doing this by myself

ckamm: If you're comfortable with the program's tracking, maker_closed_pnl / taker_closed_pnl might help? They have the pnl from reducing the position compared to avg entry price.

nicholasgodfreyclarke: I've removed `ShortToLong` and `LongToShort` here (now encapsulated in `DecreaseShort` and `DecreaseLong`).

ckamm: I'm still not getting it. (Btw, when I said "If there was no ShortToLong and LongToShort I'd kind of see the point:" I meant it in the sense of "if a single event could not decrease shorts and simultaneously increase longs..."; I don't think removing the category helps, the complexity is still there)

You said that the categories would be convenient for someone querying the fills, but I struggle to come up with usecases:

1. Query for all fills that taker-bought longs for user A
```
select * from fill_event
where taker = A
and (taker_fill_category = OpenLong 
     or taker_fill_category = IncreaseLong
     or (taker_fill_category = DecreaseShort
         and post_taker_lots > 0)
)
```
seems worse to me than
```
select * from fill_event
where taker = A
and taker_side = bid
and post_taker_lots > 0
```

2. Find all fills for user A where they taker-reduced their longs and the amount of reduction
```
select
*
, (max(0, quantity + min(post_taker_lots, 0))) as amount
from fill_event
where taker = A
and (taker_fill_category = DecreaseLong 
     or taker_fill_category = CloseLong
)
```
doesn't really do much that
```
select
*
, (max(0, quantity + min(post_taker_lots, 0))) as amount
from fill_event
where taker = A
and taker_side = ask
and amount > 0
```
doesn't also do.

3. Similar for other queries, like finding all fills where A was a taker and opened a new long.
```
select * from fill_event
where taker = A
and (taker_fill_category = OpenLong
     or taker_fill_category = DecreaseShort)
and post_taker_lots > 0
```
vs
```
select * from fill_event
where taker = A
and taker_side = bid
and post_taker_lots <= quantity
```

Do you have examples where the fill category actually makes things simpler? My core worry is that we'll write queries that _look_ simple, but actually miss the difficulty of the special cases hidden in DecreaseShort/DecreaseLong.

ckamm: Proposal: drop the categories, keep the new post_taker/maker_lots.

Alternative: Emit two events in a single fill like mc suggested, then it becomes simple - but this might need significant work in a few places and also cost lots of overhead in the instruction, so hmm

PR #697, opened 2023-08-30
ts: example script no margin group

microwavedcola1:


ts/client/scripts/mainnet-no-margin-group.ts
```diff
@@ -0,0 +1,255 @@
...
+const MIN_VAULT_TO_DEPOSITS_RATIO = 1;
+const NET_BORROWS_WINDOW_SIZE_TS = 24 * 60 * 60;
+const NET_BORROW_LIMIT_PER_WINDOW_QUOTE = toNative(0, 6).toNumber();
+
+const defaultOracleConfig = {
+  confFilter: 0.1,
+  maxStalenessSlots: null,
+};
+
+const defaultInterestRate = {
```
microwavedcola1, line 40: All rates set to 0

ts/client/scripts/mainnet-no-margin-group.ts
```diff
@@ -0,0 +1,255 @@
...
+  MB_PAYER_KEYPAIR: string;
+  GROUP_NUM: number;
+} = process.env as any;
+
+const MIN_VAULT_TO_DEPOSITS_RATIO = 1;
+const NET_BORROWS_WINDOW_SIZE_TS = 24 * 60 * 60;
+const NET_BORROW_LIMIT_PER_WINDOW_QUOTE = toNative(0, 6).toNumber();
+
+const defaultOracleConfig = {
+  confFilter: 0.1,
```
microwavedcola1, line 36: would you change this to confFilter 0?

ts/client/scripts/mainnet-no-margin-group.ts
```diff
@@ -0,0 +1,255 @@
...
+  const usdcMainnetOracle = new PublicKey(MAINNET_ORACLES.get('USDC')!);
+  let sig = await client.tokenRegister(
+    group,
+    usdcMainnetMint,
+    usdcMainnetOracle,
+    defaultOracleConfig,
+    0,
+    'USDC',
+    defaultInterestRate,
+    0.0,
```
microwavedcola1, line 115: all fees and weights set to 0

ts/client/scripts/mainnet-no-margin-group.ts
```diff
@@ -0,0 +1,255 @@
...
+  MB_CLUSTER_URL,
+  MB_PAYER_KEYPAIR,
+  GROUP_NUM,
+}: {
+  MB_CLUSTER_URL: string;
+  MB_PAYER_KEYPAIR: string;
+  GROUP_NUM: number;
+} = process.env as any;
+
+const MIN_VAULT_TO_DEPOSITS_RATIO = 1;
```
microwavedcola1, line 31: No borrowing allowed

ts/client/scripts/mainnet-no-margin-group.ts
ckamm: Can't disable the conf filter right now. The logic is `if deviation > conf_filter * price { error }`, so make it big without overflowing? 1000?

ts/client/scripts/mainnet-no-margin-group.ts
ckamm: Since utilization will always be 0, rates should be 0 anyway. I think we need to check that the fast-listing instruction is usable? Could be, I think it sets reduce_only=2

ckamm:
Ok, so liquidator triggering needs to be fixed (oracles should update quickly enough that waiting for them is kind of pointless) and to not depend on a serum market.

microwavedcola1:
Cant register usdc as trustless, which is ok I guess, see `require_neq!(token_index, QUOTE_TOKEN_INDEX);`

PR #698, opened 2023-08-30
Account size migration

ckamm:
- shrink accounts where possible
- add reserved space at the end of the dynamic part of a mango account,
  to enable future expansion with less headache


ckamm:
Will fold into the account resizing PR

PR #705, opened 2023-09-05
Ts/ids fixes

tlrsssss:
Using the client in 'api' mode isn't working because the Ids class doesn't expose an 'active' flag available for many of the banks, markets, etc

Deprecate ids.json. I don't think it ever worked and doesn't make sense now. Also deprecate the functions that rely on it.

### DIFF
--- a/ts/client/src/accounts/group.ts
+++ b/ts/client/src/accounts/group.ts
@@ -144,7 +144,7 @@ export class Group {
       ),
       this.reloadMintInfos(client, ids),
       this.reloadSerum3Markets(client, ids).then(() =>
-        this.reloadSerum3ExternalMarkets(client),
+        this.reloadSerum3ExternalMarkets(client, ids),
       ),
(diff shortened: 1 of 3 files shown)


ts/client/src/accounts/group.ts
```diff
@@ -280,23 +280,68 @@ export class Group {
...
-          serum3Market.serumMarketExternal,
-          { commitment: client.program.provider.connection.commitment },
-          OPENBOOK_PROGRAM_ID[client.cluster],
+  public async reloadSerum3ExternalMarkets(
+    client: MangoClient,
+    ids?: Id,
+  ): Promise<void> {
+    let markets: Market[] = [];
+    const externalMarketIds = ids?.getSerum3ExternalMarkets();
+    console.log(1, externalMarketIds);
```
microwavedcola1, line 289: nitpick: remove this and other console logs

PR #706, opened 2023-09-07
Mc/tcs improvements

microwavedcola1:


ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1973,6 +1959,62 @@ export class TokenConditionalSwap {
...
+    const buyBank = this.getBuyToken(group);
+    const sellBank = this.getSellToken(group);
+    return TokenConditionalSwap.computePremium(
+      group,
+      buyBank,
+      sellBank,
+      this.maxBuy,
+      this.maxSell,
+      this.getMaxBuyUi(group),
+      this.getMaxSellUi(group),
```
ckamm, line 1972: Why pass maxBuy and maxBuyUi if the bank is also passed? 

For the currently suggested premium, it could make sense to use the remaining buy/sell amounts, not the original maximums. Maybe it's half-filled already.

microwavedcola1: > Why pass maxBuy and maxBuyUi if the bank is also passed?
Fixed

microwavedcola1: Fixed both

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1973,6 +1959,61 @@ export class TokenConditionalSwap {
...
 
+  getCurrentlySuggestedPremium(group: Group): number {
+    const buyBank = this.getBuyToken(group);
+    const sellBank = this.getSellToken(group);
+    return TokenConditionalSwap.computePremium(
+      group,
+      buyBank,
+      sellBank,
+      this.maxBuy.sub(this.bought),
+      this.maxSell.sub(this.sold),
```
ckamm, line 1970: Ah, I think I made everything more difficult here. You rely on the special case with U64_MAX_BN in computePremium, which subtracting here would break.

Either roll this back and just pass maxBuy/maxSell or also pass bought/sold and do the subtraction in the function.

microwavedcola1: rolled back

ckamm (approved):

PR #711, opened 2023-09-10
Mc/token withdraw all deposit for mint

microwavedcola1:


ts/client/src/client.ts
```diff
@@ -1409,6 +1409,35 @@ export class MangoClient {
...
+    const bank = group.getFirstBankByMint(mintPk);
+    const b = mangoAccount.getTokenBalance(bank).toNumber();
+    if (b < 0) {
+      throw new Error(`Only call this method for deposits and not borrows!`);
+    }
+    const ixes = await this.tokenWithdrawNativeIx(
+      group,
+      mangoAccount,
+      mintPk,
+      new BN(b),
```
ckamm, line 1434: I recommend sending u64::MAX here. Otherwise some interest may accrue between this call and the on-chain execution.

PR #715, opened 2023-09-11
Update mango 0.18 to solana 1.16 + anchor 28 - Part I of update

grooviegermanikus:
* use anchor 0.28.0 (was 0.27.0)
* use solana 1.16.13 (>=1.16.7, was 1.14.9)
  * 1.16.7 is the first version that supports borsh 0.10 [PR](https://github.com/solana-labs/solana/pull/32511)
* aligned borsh 0.9 + borsh 0.10
* use different version of _fixed_ library
* updated to switchboard sbv2-crate-0.4.0
* reqwest fails with 0.11.18 - sticking to 0.11.17
* note: i have blindy updated the version number in ci-code-review-rust.yml + ci-soteria.yml 
* **TODO** decide which mango-v4 version to use
* **TODO** update to mango-v4 0.19

See also the [mango-feeds PR](https://github.com/blockworks-foundation/mango-feeds/pull/16)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,14 +12,14 @@ anchor-spl = { git = "https://github.com/coral-xyz/anchor.git", rev = "2af9cc669
 fixed = { git = "https://gitlab.com/ckamm/fixed.git", branch = "v1.11.0-mango" }
 pyth-sdk-solana = "0.7.0"
 serum_dex = { git = "https://github.com/openbook-dex/program.git" }
-solana-address-lookup-table-program = "~1.14.9"
-solana-account-decoder = "~1.14.9"
-solana-client = "~1.14.9"
-solana-logger = "~1.14.9"
-solana-program = "~1.14.9"
-solana-program-test = "~1.14.9"
-solana-rpc = "~1.14.9"
-solana-sdk = "~1.14.9"
+solana-address-lookup-table-program = "~1.14.18"
+solana-account-decoder = "~1.14.18"
+solana-client = "~1.14.18"
+solana-logger = "~1.14.18"
+solana-program = "~1.14.18"
+solana-program-test = "~1.14.18"
+solana-rpc = "~1.14.18"
+solana-sdk = "~1.14.18"
 
 [profile.release]
 overflow-checks = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ anchor-client = { git = "https://github.com/coral-xyz/anchor.git", rev = "2af9cc
 anchor-lang = { git = "https://github.com/coral-xyz/anchor.git", rev = "2af9cc669", version = "0.28.0" }
 anchor-spl = { git = "https://github.com/coral-xyz/anchor.git", rev = "2af9cc669", version = "0.28.0" }
 # use ckamm commit 2363acc4 (v1.23.0-mango -1)
-fixed = { git = "https://gitlab.com/ckamm/fixed.git", rev = "2363acc4" }
+fixed = { git = "https://github.com/blockworks-foundation/fixed.git", branch = "v1.11.0-borsh0_10-mango" }
 pyth-sdk-solana = "0.8.0"
 serum_dex = { git = "https://github.com/openbook-dex/program.git" }
 # 1.16.7+ is required due to this: https://github.com/blockworks-foundation/mango-v4/issues/712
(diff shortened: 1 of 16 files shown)


ckamm:
Great, will need a bit to check.

Upgrading "fixed" to 1.23.0 is definitely not good. That version had a critical bug (https://gitlab.com/tspiteri/fixed/-/issues/57) that got fixed in 1.23.1. We downgraded to tried-and-true 1.11.0 for that reason.

I assume you upgraded due to the borsh dependency. I think I'd be in favor of backporting the fixed borsh upgrade into 1.11.0.

programs/mango-v4/Cargo.toml
```diff
@@ -29,7 +29,7 @@ anchor-lang = { workspace = true }
 anchor-spl = { workspace = true }
 arrayref = "0.3.6"
 bincode = "1.3.3"
-borsh = { version = "0.9.3", features = ["const-generics"] }
+borsh = { version = "0.10.3", features = ["const-generics"] }
```
ckamm, line 32: fwiw, I think const-generics support is default in borsh 0.10, can drop the features specification

programs/mango-v4/Cargo.toml
```diff
@@ -44,7 +44,7 @@ solana-sdk = { workspace = true, default-features = false, optional = true }
 solana-security-txt = "1.1.0"
 static_assertions = "1.1"
 switchboard-program = ">=0.2.0"
-switchboard-v2 = "0.1.17"
+switchboard-v2 = { git = "https://github.com/switchboard-xyz/solana-sdk.git", tag = "sbv2-crate-0.4.0" }
```
ckamm, line 47: switchboard-v2 seems to have the 0.4.0 crate published - why do we need the git dependency?

grooviegermanikus: on [crates.io](https://crates.io/crates/switchboard-v2/0.4.0) it says:
> Deprecated in favor of the switchboard-solana crate.
this seems to offer a published version for anchor 28 - will try with that one again

grooviegermanikus: using this now:
`switchboard-v2 = { package = "switchboard-solana", version = "0.28.18" }`

Cargo.toml
```diff
@@ -6,20 +6,23 @@ members = [
 ]
 
 [workspace.dependencies]
-anchor-client = "0.27.0"
-anchor-lang = "0.27.0"
-anchor-spl = "0.27.0"
-fixed = { path = "./3rdparty/fixed", version = "1.11.0" }
-pyth-sdk-solana = "0.7.0"
+# rev 2af9cc669 is successor of 0.28.0
+anchor-client = { git = "https://github.com/coral-xyz/anchor.git", rev = "2af9cc669", version = "0.28.0" }
```
godmodegalactus, line 10: should use 0.28.0 instead

godmodegalactus (changes requested):

Cargo.toml
ckamm: Really? Wasn't there a good reason to go to 2af9cc669, due to solana version issues with 0.28.0?

acheroncrypto: Anchor `0.28.0` should work just fine here since Solana crates are also being updated. https://github.com/coral-xyz/anchor/pull/2572 was only for `solana-client 1.14`.

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax = docker/dockerfile:1.2
 # Base image containing all binaries, deployed to ghcr.io/blockworks-foundation/mango-v4:latest
-FROM rust:1.65 as base
+FROM rust:1.70.0-bullseye as base
```
godmodegalactus, line 3: should be rust:1.70.0

grooviegermanikus: but why? we rely on Debian image later

Cargo.toml
grooviegermanikus: thx will do

grooviegermanikus: tried with vanilla anchor 0.28 but that gives me this error:

> 39 |     pub openbook_v2_market_external: AccountLoader<'info, Market>,
>    |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `ZeroCopy` is not implemented for `openbook_v2::state::Market`

you can try by comparing e9b23d587 and fa6e24c90

godmodegalactus: openbook_v2::market has been defined as zero copy. Is the branch latest branch ?

grooviegermanikus:
will continue on https://github.com/blockworks-foundation/mango-v4/pull/718

PR #717, opened 2023-09-13
Tcs: linear and premium auctions

ckamm:


programs/mango-v4/src/state/token_conditional_swap.rs
```diff
@@ -148,10 +194,22 @@ impl TokenConditionalSwap {
...
     pub fn is_expired(&self, now_ts: u64) -> bool {
         now_ts >= self.expiry_timestamp
     }
 
+    pub fn passed_start(&self, now_ts: u64) -> bool {
+        self.start_timestamp > 0 && now_ts >= self.start_timestamp
+    }
+
+    pub fn has_incentive_for_starting(&self) -> bool {
+        self.tcs_type() == TokenConditionalSwapType::PremiumAuction
```
microwavedcola1, line 210: hmm so the linear auction has no incentive to start?

programs/mango-v4/src/state/token_conditional_swap.rs
```diff
@@ -197,10 +290,74 @@ impl TokenConditionalSwap {
...
+            TokenConditionalSwapType::FixedPremium => {
+                require!(
+                    self.price_in_range(price),
+                    MangoError::TokenConditionalSwapPriceNotInRange
+                );
+            }
+            TokenConditionalSwapType::PremiumAuction => {
+                // Technically the error is lying, since it's also triggerable once started
+                require!(
+                    self.price_in_range(price) || self.passed_start(now_ts),
```
microwavedcola1, line 343: why is this check not simply ` self.passed_start(now_ts)`? the `self.price_in_range(price)` would be implicitly taken care once the tcs start ix for premium auction is called

programs/mango-v4/src/state/token_conditional_swap.rs
```diff
@@ -168,11 +226,46 @@ impl TokenConditionalSwap {
...
+        match self.tcs_type() {
+            TokenConditionalSwapType::FixedPremium => base_price * (1.0 + self.price_premium_rate),
+            TokenConditionalSwapType::PremiumAuction => {
+                // Start dynamically when triggerable
+                let start = self.start_timestamp_or_now(now_ts);
+                assert!(start <= now_ts);
+
+                let duration = self.duration_seconds as f64;
+                let current = (now_ts - start) as f64;
+                let progress = (current / duration).min(1.0);
```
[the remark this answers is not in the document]
microwavedcola1: oh, dont you want max instead?

programs/mango-v4/src/state/token_conditional_swap.rs
ckamm: Linear auctions always start at a pre-defined start_timestamp. There's no need to call start() once a condition is met.

programs/mango-v4/src/state/token_conditional_swap.rs
ckamm: I want to allow triggering directly without start ever being called.

programs/mango-v4/src/instructions/token_conditional_swap_start.rs
```diff
@@ -0,0 +1,141 @@
...
+#[allow(clippy::too_many_arguments)]
+pub fn token_conditional_swap_start(
+    ctx: Context<TokenConditionalSwapStart>,
+    token_conditional_swap_index: usize,
+    token_conditional_swap_id: u64,
+) -> Result<()> {
+    let group_pk = &ctx.accounts.group.key();
+    let account_key = ctx.accounts.account.key();
+    let caller_key = ctx.accounts.caller.key();
+    require_keys_neq!(account_key, caller_key);
```
microwavedcola1, line 21: cant this check be moved to the accounts mod? where the context is defined

programs/mango-v4/src/instructions/token_conditional_swap_start.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    let mut account = ctx.accounts.account.load_full_mut()?;
```
microwavedcola1, line 23: why dont we also check if caller is not being liquidated?

programs/mango-v4/src/instructions/token_conditional_swap_start.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    let mut caller = ctx.accounts.caller.load_full_mut()?;
+
+    let mut account_retriever = ScanningAccountRetriever::new(ctx.remaining_accounts, group_pk)
+        .context("create account retriever")?;
+
+    let tcs = account
+        .token_conditional_swap_by_index(token_conditional_swap_index)?
+        .clone();
+    require!(tcs.has_data(), MangoError::TokenConditionalSwapNotSet);
```
microwavedcola1, line 37: can we rename the has_data to is_active? that would be consistent with other things we have no? well the only difference is that it can be expired I guess, but has_data seems to low level imo

ckamm: I agree about this being inconsistent with other components, but `is_active` seemd confusing to me: It could be expired or it could be an auction that hasn't started yet.

`is_set`, `is_configured`, `is_defined`?

programs/mango-v4/src/instructions/token_conditional_swap_start.rs
```diff
@@ -0,0 +1,141 @@
...
+        incentive_token_index: sell_token_index,
+        incentive_amount: incentive_native,
+    });
+
+    //
+    // Start the tcs
+    //
+    let tcs = account.token_conditional_swap_mut_by_index(token_conditional_swap_index)?;
+    tcs.start_timestamp = now_ts;
+    tcs.sold += incentive_native;
```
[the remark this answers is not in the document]
ckamm: My thoughts on this were that the incentive should come out of the sell-token pool, otherwise you could create premium auctions for tiny sell amounts where the incentive dwarves the actual amount sold.

Otoh, right now it's possible to create auctions that get completely filled by the incentive payment.

How about I forbid creating premium auctions where the sell token amount is <10x the incentive?

programs/mango-v4/src/instructions/token_conditional_swap_start.rs
microwavedcola1: is_configured lgtm

programs/mango-v4/src/instructions/token_conditional_swap_start.rs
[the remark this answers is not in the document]
microwavedcola1: > How about I forbid creating premium auctions where the sell token amount is <10x the incentive?

yes please

PR #719, opened 2023-09-15
release 19.1 -> deploy + serum3 open orders estimation ts patch

microwavedcola1:


ckamm (approved):

programs/mango-v4/src/health/cache.rs
```diff
@@ -258,7 +294,13 @@ impl Serum3Info {
         let quote_asset = quote_info.prices.asset(health_type);
         let base_liab = base_info.prices.liab(health_type);
         // OPTIMIZATION: These divisions can be extremely expensive (up to 5k CU each)
-        self.reserved_base + self.reserved_quote * quote_asset / base_liab
+        let reserved_quote_as_base_oracle = self.reserved_quote * quote_asset / base_liab;
```
mschneider, line 297: what would be the impact of calculating this in f64, given that reserved_quote_as_base_highest_bid is stored in f64 anyways?

programs/mango-v4/src/instructions/serum3_place_order.rs
```diff
@@ -233,24 +167,65 @@ pub fn serum3_place_order(
...
+        serum.lowest_placed_ask = 0.0;
+    }
+    let new_order_on_book = after_oo_free_slots != before_oo_free_slots;
+    if new_order_on_book {
+        match side {
+            Serum3Side::Ask => {
+                // in the normal quote per base units
+                let limit_price =
+                    limit_price_lots as f64 * quote_lot_size as f64 / base_lot_size as f64;
+                serum.lowest_placed_ask = if serum.lowest_placed_ask == 0.0 {
```
mschneider, line 198: why is 0 the reserved value instead of something impossible to input: e.g. NaN?

mschneider: what is the impact of placing one big ask at $0 and a small one at $100?

programs/mango-v4/src/health/cache.rs
ckamm: I suspect that f64 computation would be fine. The error in the oracle price should way bigger than any rounding. Measuring and testing this had been a possibility (hence the "// optimization" comments), but serum3 health contributions haven't been a limiting compute factor so far.

programs/mango-v4/src/instructions/serum3_place_order.rs
ckamm: Zero is the reserved value for migration reasons: All existing Serum3Orders structs will have 0.0 in these fields, so we need to special case for it anyway.

You can't place orders at 0. But if you placed an ask at $0.001 or something and another at $100 then both reserved base amounts would be valued at $0.001. This is an intentional underestimation we make because we can't track the true prices and quantities.

mschneider (approved):

programs/mango-v4/src/instructions/serum3_place_order.rs
mschneider: gotcha

PR #725, opened 2023-09-19
Mc/settler cu limit

microwavedcola1:


lib/client/src/client.rs
```diff
@@ -1809,12 +1809,13 @@ impl TransactionBuilder {
...
     ) -> anyhow::Result<solana_sdk::transaction::VersionedTransaction> {
         let latest_blockhash = rpc.get_latest_blockhash().await?;
-        self.transaction_with_blockhash(latest_blockhash)
+        self.transaction_with_blockhash(latest_blockhash, Option::None)
     }
 
     pub fn transaction_with_blockhash(
         &self,
         blockhash: Hash,
+        compute_unit_limit_opt: Option<u32>,
```
ckamm, line 1818: instead of adding it as an optional argument, how about adding a `prepend_compute_limit_instruction()` function to TransactionBuilder?

lib/client/src/client.rs
```diff
@@ -1825,6 +1826,15 @@ impl TransactionBuilder {
                 ),
             )
         }
+        let mut ix = self.instructions.clone();
```
ckamm, line 1829: this completely hides the previous `ix`, dropping the compute budget instruction added before

ckamm (changes requested):

lib/client/src/client.rs
microwavedcola1: how about doing this instead

image

ckamm: I'd like to keep client specific logic out of `TransactionBuilder`. How about this:
- Add `default_compute_budget_per_instruction` field to `TransactionBuilderConfig`. The client sets this to Some(250k) by default
- In transaction_with_blockhash: check if a compute budget / compute unit price ix exists, if not, add it with the values from TransactionBuilderConfig.

ckamm: pushed fixup

PR #726, opened 2023-09-20
ts: helpers for withdrawing tokens from bad oracles

microwavedcola1:


ts/client/src/accounts/mangoAccount.ts
```diff
@@ -182,6 +182,17 @@ export class MangoAccount {
     return this.frozenUntil.lt(new BN(Date.now() / 1000));
   }
 
+  public async tokenPositionsForNotConfidentOrStaleOracles(
```
microwavedcola1, line 185: So I am thinking like this, the UI basically calls this, and knows which banks have deposits, and which have borrows, 

for borrows - it asks user to close borrow
for deposits - it asks user to withdraw all using the new helper in client.ts

ts/client/src/accounts/bank.ts
```diff
@@ -64,6 +64,7 @@ export class Bank implements BankForHealth {
   public _price: I80F48 | undefined;
   public _uiPrice: number | undefined;
   public _oracleLastUpdatedSlot: number | undefined;
+  public _oracleLastKnowndeviation: I80F48 | undefined;
```
ckamm, line 67: capitalize D

ts/client/src/accounts/group.ts
```diff
@@ -472,11 +470,23 @@ export class Group {
...
       lastUpdatedSlot = stubOracle.lastUpdateSlot.toNumber();
       provider = OracleProvider.Stub;
+      deviation = stubOracle.deviation;
     } else if (isPythOracle(ai)) {
       const priceData = parsePriceData(ai.data);
       uiPrice = priceData.previousPrice;
       price = this.toNativePrice(uiPrice, baseDecimals);
       lastUpdatedSlot = parseInt(priceData.lastSlot.toString());
+      deviation = priceData.confidence
+        ? I80F48.fromNumber(
```
ckamm, line 480: I think you want to deal with `undefined` but for confidence == 0 we should set deviation to 0 too. So `confidence !== undefined`?

ts/client/src/accounts/oracle.ts
```diff
@@ -133,3 +155,26 @@ export function isSwitchboardOracle(accountInfo: AccountInfo<Buffer>): boolean {
 export function isPythOracle(accountInfo: AccountInfo<Buffer>): boolean {
   return accountInfo.data.readUInt32LE(0) === PythMagic;
 }
+
+export function checkOracleConfidenceAndStaleness(
```
ckamm, line 159: It's surprising to me that this is true when it's stale or inconfident. Could this be `isOracleStaleOrUncofident()`? Or reverse the output and use `isOracleHealthy` or something

ts/client/src/accounts/group.ts
```diff
@@ -472,11 +470,23 @@ export class Group {
...
+            priceData.confidence *
+              Math.pow(
+                10,
+                priceData.exponent +
```
ckamm, line 484: did you verify this one? priceData.previousPrice seems to already have the exponent factored in, so it's odd that this needs it to be applied manually

ts/client/src/accounts/oracle.ts
```diff
@@ -63,26 +64,47 @@ export class StubOracle {
...
+  mantissa: BN;
+  scale: number;
+}): Big {
+  const mantissa = new Big(sbDecimal.mantissa.toString());
+  const scale = sbDecimal.scale;
+  const oldDp = Big.DP;
+  Big.DP = 20;
+  const result: Big = mantissa.div(new Big(10).pow(scale));
+  Big.DP = oldDp;
+  return result;
```
ckamm, line 90: where's this code from? I've compared it to rusts's Decimal::to_f64 - is that the right place?

ts/client/src/accounts/oracle.ts
microwavedcola1: changed to isOracleStaleOrUncofident

ts/client/src/accounts/oracle.ts
microwavedcola1: https://github.com/switchboard-xyz/solana-sdk/blob/7eeab51398a67abe08999ddc5c6b5e693537c1cb/javascript/sbv2-lite/src/index.ts#L173

Its picked up from here

ts/client/src/accounts/group.ts
```diff
@@ -472,11 +470,19 @@ export class Group {
...
+      deviation = stubOracle.deviation;
     } else if (isPythOracle(ai)) {
       const priceData = parsePriceData(ai.data);
       uiPrice = priceData.previousPrice;
       price = this.toNativePrice(uiPrice, baseDecimals);
       lastUpdatedSlot = parseInt(priceData.lastSlot.toString());
+      deviation =
+        priceData.confidence !== undefined
+          ? I80F48.fromNumber(
+              priceData.previousConfidence *
```
ckamm, line 482: since previousConfidence is always a number, can we replace all of this with `deviation = this.toNativePrice(priceData.previousConfidence, baseDecimals)` to parallel the `price`?

ts/client/src/accounts/oracle.ts
```diff
@@ -133,3 +155,26 @@ export function isSwitchboardOracle(accountInfo: AccountInfo<Buffer>): boolean {
 export function isPythOracle(accountInfo: AccountInfo<Buffer>): boolean {
   return accountInfo.data.readUInt32LE(0) === PythMagic;
 }
+
+export function isOracleStaleOrUncofident(
```
ckamm: -> Unconfident

PR #728, opened 2023-09-20
Mc/cu opt

microwavedcola1:


ts/client/src/utils/rpc.ts
```diff
@@ -45,15 +45,18 @@ export async function sendTransaction(
...
-      ) && u8().decode(ix.data.subarray(0, 1)) == 2,
+      ix.programId.equals(COMPUTE_BUDGET_PROGRAM_ID) &&
+      u8().decode(ix.data.subarray(0, 1)) == 2,
   );
 
   if (!computeUnitLimitIxFound) {
+    const totalUserIntendedIxs = ixs.filter(
+      (ix) => !ix.programId.equals(COMPUTE_BUDGET_PROGRAM_ID),
+    ).length;
+    const requestCu = Math.min(totalUserIntendedIxs * 250_000, 1_600_000);
```
microwavedcola1, line 56: I don't think I should make special case for health region been added, correct?

ckamm (approved):

PR #733, opened 2023-09-28
ts: scripts for testing account shrink

microwavedcola1:


ts/client/src/client.ts
```diff
@@ -1533,6 +1532,28 @@ export class MangoClient {
...
+    );
+    const eq = decodeEventQueue(ai!.data);
+    const orderedAccounts: PublicKey[] = eq
+      .map((e) => e.openOrders)
+      .sort((a, b) => a.toBuffer().swap64().compare(b.toBuffer().swap64()));
+    if (orderedAccounts.length == 0) {
+      throw new Error(`Event queue is empty!`);
+    }
+    return this.sendAndConfirmTransactionForGroup(group, [
+      serum3MarketExternal.makeConsumeEventsInstruction(orderedAccounts, 65535),
```
ckamm, line 1553: This will exceed cu limits if too many events are processed at a time, what that something you accepted? I suggest we
- only use accounts for the top 10 events (instead of all, which could break the limit)
- send a limit of 10

ckamm:
@microwavedcola1 merge or close?

PR #735, opened 2023-09-29
ts: upgrade anchor

microwavedcola1:


ckamm (approved):

ts/client/src/client.ts
```diff
@@ -1855,11 +1855,11 @@ export class MangoClient {
     group: Group,
     mangoAccount: MangoAccount,
     externalMarketPk: PublicKey,
-    side: Serum3Side,
+    side: Serum3Side | any,
```
microwavedcola1, line 1858: image

So basically we would have to move away from current to globals, see both current and global example below
image

microwavedcola1: I think my current workaround is better, we keep code as is, but have a `| any` so the caller knows the main type he should use, and code compiles

microwavedcola1: Alternatively we could disable compile checking completely for these lines, or do you have other suggestions?

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -1793,15 +1793,29 @@ export class PerpOoDto {
   ) {}
 }
 
-export class TokenConditionalSwapDisplayPriceStyle {
-  static sellTokenPerBuyToken = { sellTokenPerBuyToken: {} };
-  static buyTokenPerSellToken = { buyTokenPerSellToken: {} };
+export type TokenConditionalSwapDisplayPriceStyle =
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  | { sellTokenPerBuyToken: {} }
```
ckamm, line 1798: I suggest using the recommended `Record<string, none>` over `{}` here so we can get rid of the lint-comment

ckamm (approved):

PR #745, opened 2023-10-09
helpers for grouping and spotting tx errors

microwavedcola1:


ts/client/src/error.ts
```diff
@@ -0,0 +1,42 @@
...
+export enum WellKnownTransactionErrors {
+  // Slippage incurred was higher than user expected
+  JupiterSlippageToleranceExceeded,
+  Unknown,
+}
+
+export function isAJupiterTx(logMessages: string[]): boolean {
+  return (
+    logMessages.filter((msg) => msg.indexOf(JUPITER.V3.toBase58()) > -1)
+      .length > 0 ||
```
ckamm, line 13: Simplifications:
- `logMessages.filter(pred).length > 0` -> `logMessages.some(pred)`
- `msg.indexOf(str) > -1` -> `msg.includes(str)`

ts/client/src/error.ts
```diff
@@ -0,0 +1,42 @@
+import { Connection } from '@solana/web3.js';
+import { JUPITER } from './constants';
+
+export enum WellKnownTransactionErrors {
```
ckamm, line 4: Could this just be `TransactionErrors`?

ts/client/src/error.ts
```diff
@@ -0,0 +1,42 @@
...
+  // Slippage incurred was higher than user expected
+  JupiterSlippageToleranceExceeded,
+  Unknown,
+}
+
+export function isAJupiterTx(logMessages: string[]): boolean {
```
ckamm, line 10: maybe `containsJupiterProgram()` to be more technically correct?

ckamm (approved):

PR #752, opened 2023-10-11
gov-ix: create an idl-resize instruction

ckamm:


ts/client/scripts/create-gov-ix.ts
```diff
@@ -381,15 +382,45 @@ async function createMangoAccount(): Promise<void> {
   console.log(await serializeInstructionToBase64(ix));
 }
 
+async function idlResize(): Promise<void> {
+  const idlIxBytes = [0x40, 0xf4, 0xbc, 0x78, 0xa7, 0xe9, 0x69, 0x0a];
```
microwavedcola1, line 386: nit: can you link to some code where you grabbed these? or was this from some real txs?

microwavedcola1 (approved):

ts/client/scripts/create-gov-ix.ts
ckamm: that's just the anchor constant for idl instruction calls, you can find it in the anchor source or in instructions that set idl etc

PR #775, opened 2023-11-07
program: even delegate can create token position

microwavedcola1:


programs/mango-v4/src/accounts_ix/token_deposit.rs
```diff
@@ -53,10 +53,12 @@ pub struct TokenDeposit<'info> {
...
         mut,
         has_one = group,
-        has_one = owner,
-        constraint = account.load()?.is_operational() @ MangoError::AccountIsFrozen
+        constraint = account.load()?.is_operational() @ MangoError::AccountIsFrozen,
+        constraint = account.load()?.is_owner_or_delegate(token_authority.key()),
     )]
     pub account: AccountLoader<'info, MangoAccountFixed>,
+    // Unused, check is now on the token authority instead
+    // we want to support even delegates opening new token positions
```
ckamm, line 61: Stopping to support token_authority != owner would be a breaking change! (even though I doubt anyone uses it). Let's instead just check `is_owner_or_delegate(owner.key())`?

ckamm (approved):

PR #780, opened 2023-11-14
Bank: allow maint weights to change over time

ckamm:
- token_edit can set it up to gradually scale to new target values
- security admin can abort an ongoing change via token_edit
- all health computations are now time dependent and get the weight based on it
- when the change is done, the keeper "cleans up" and moves the new values into the default fields

### DIFF
--- a/programs/mango-v4/src/instructions/token_edit.rs
+++ b/programs/mango-v4/src/instructions/token_edit.rs
@@ -44,6 +44,11 @@ pub fn token_edit(
     flash_loan_swap_fee_rate_opt: Option<f32>,
     interest_curve_scaling_opt: Option<f32>,
     interest_target_utilization_opt: Option<f32>,
+    maint_weight_shift_start_opt: Option<u64>,
+    maint_weight_shift_end_opt: Option<u64>,
+    maint_weight_shift_asset_target_opt: Option<f32>,
+    maint_weight_shift_liab_target_opt: Option<f32>,
+    maint_weight_shift_abort: bool,
 ) -> Result<()> {
     let group = ctx.accounts.group.load()?;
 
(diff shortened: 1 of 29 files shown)


microwavedcola1 (approved):

programs/mango-v4/src/instructions/token_edit.rs
```diff
@@ -44,6 +44,11 @@ pub fn token_edit(
     flash_loan_swap_fee_rate_opt: Option<f32>,
     interest_curve_scaling_opt: Option<f32>,
     interest_target_utilization_opt: Option<f32>,
+    maint_weight_shift_start_opt: Option<u64>,
+    maint_weight_shift_end_opt: Option<u64>,
+    maint_weight_shift_asset_target_opt: Option<f32>,
```
microwavedcola1, line 49: hmm why is the maint_weight_shift_duration_inv not taken here?

programs/mango-v4/src/state/bank.rs
```diff
@@ -161,8 +161,14 @@ pub struct Bank {
     // can be negative due to multibank, then it'd need to be balanced in the keeper
     pub deposits_in_serum: i64,
 
+    pub maint_weight_shift_start: u64,
+    pub maint_weight_shift_end: u64,
+    pub maint_weight_shift_duration_inv: I80F48,
```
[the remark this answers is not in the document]
microwavedcola1: ok I see its simply a toggle, so should we keep it linear or non-linear?

programs/mango-v4/src/instructions/token_edit.rs
microwavedcola1: ok I see, its simply a toggle

programs/mango-v4/src/state/bank.rs
```diff
@@ -338,6 +354,26 @@ impl Bank {
...
+        {
+            (self.maint_asset_weight, self.maint_liab_weight)
+        } else if now_ts >= self.maint_weight_shift_end {
+            (
+                self.maint_weight_shift_asset_target,
+                self.maint_weight_shift_liab_target,
+            )
+        } else {
+            let scale = I80F48::from(now_ts - self.maint_weight_shift_start)
+                * self.maint_weight_shift_duration_inv;
```
microwavedcola1, line 368: any arguments for non-linear movement?

microwavedcola1 (approved):

PR #785, opened 2023-11-19
Reduce HealthCache heap use

ckamm:
Growing these vectors is a waste, rough measurements say this saves around 1k bytes for a 5 tokens and 4 serum account.


mschneider (approved):
LGTM

PR #786, opened 2023-11-20
program: optimize cu usage in token deposit, by computing health only…

microwavedcola1:
… if user is being liquidated

### DIFF
--- a/programs/mango-v4/src/instructions/token_deposit.rs
+++ b/programs/mango-v4/src/instructions/token_deposit.rs
@@ -129,11 +129,14 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
             msg!("health: {}", health);
 
             let was_being_liquidated = account.being_liquidated();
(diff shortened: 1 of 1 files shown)


programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -124,11 +124,12 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
...
+        let was_being_liquidated = account.being_liquidated();
+        if !account.fixed.is_in_health_region() && was_being_liquidated {
             let health = cache.health(HealthType::LiquidationEnd);
             msg!("health: {}", health);
+            // Only compute health and check for recovery if not already being liquidated
 
-            let was_being_liquidated = account.being_liquidated();
             let recovered = account.fixed.maybe_recover_from_being_liquidated(health);
             require!(
                 !was_being_liquidated || recovered,
```
ckamm, line 135: nit: !was_being_liquidated is always false here, can drop it

ckamm (approved):

PR #790, opened 2023-11-20
Add fallback oracles

Lou-Kamades:


Lou-Kamades:
Before implementing the typescript and rust client code for this change I wanted to ask:

1. Are we ok with using remaining accounts in this way?
2. Do we want to limit which instructions can access fallback oracles? If so, which ones?  Apart from ensuring that only certain instructions will check for fallback oracles, we can also control/test this feature by only adding a fallback for certain coins

lib/client/src/health_cache.rs
```diff
@@ -34,6 +34,7 @@ pub async fn new(
         begin_perp: active_token_len * 2,
         begin_serum3: active_token_len * 2 + active_perp_len * 2,
         staleness_slot: None,
+        fallback_oracle_ais: vec![],
```
ckamm, line 37: This poses an interesting question: How do clients deal with fallback oracles?

Processes like the liquidator that have no trouble fetching a few more oracle accounts could grab the fallback accounts, compute staleness client side and then pass in up to a configurable number of fallbacks if needed.

It'll need a bunch of work and this function is one of the places that would be affected.

Some code, like the "is this liquidatable?" check, would want to pass in all fallbacks so we never give up too early. Then presumably later code would figure out what fallbacks are actually needed?

None of this has to happen in this PR, but maybe mark this line as needing to be revisited.

programs/mango-v4/src/instructions/token_edit.rs
```diff
@@ -49,6 +49,7 @@ pub fn token_edit(
     maint_weight_shift_asset_target_opt: Option<f32>,
     maint_weight_shift_liab_target_opt: Option<f32>,
     maint_weight_shift_abort: bool,
+    fallback_oracle_opt: Option<Pubkey>,
```
ckamm, line 52: This passes the oracle twice, as an account and as an instruction argument. Doesn't seem to be necessary. Instead just pass a `set_fallback_oracle: bool`? (we already should have caught this for the main oracle)

programs/mango-v4/src/state/bank.rs
```diff
@@ -292,7 +294,8 @@ impl Bank {
             maint_weight_shift_duration_inv: existing_bank.maint_weight_shift_duration_inv,
             maint_weight_shift_asset_target: existing_bank.maint_weight_shift_asset_target,
             maint_weight_shift_liab_target: existing_bank.maint_weight_shift_liab_target,
-            reserved: [0; 2008],
+            fallback_oracle: existing_bank.oracle, // bongo
```
ckamm, line 297: bingo? ;)

programs/mango-v4/src/state/bank.rs
```diff
@@ -955,6 +966,23 @@ impl Bank {
...
 
+    /// Tries to return the primary oracle price, and if there is a confidence or staleness issue returns the fallback oracle price.
+    pub fn oracle_price_with_fallback(
+        &self,
+        oracle_acc: &impl KeyedAccountReader,
+        fallback_oracle_acc_opt: Option<&impl KeyedAccountReader>,
+        staleness_slot: Option<u64>,
+    ) -> Result<I80F48> {
+        let primary_price = self.oracle_price(oracle_acc, staleness_slot);
+        if primary_price.is_ok() || fallback_oracle_acc_opt.is_none() {
```
ckamm, line 977: Users could pass the wrong account for `oracle_acc` here, then the fallback would be used independently of the state of the main oracle!

There's different ways to go about this: I think it's appropriate here to check that the actual error returned is one of the ones that should enable going to a fallback oracle.

Note also that check_confidence_and_maybe_staleness() does some logging on failure that might be inappropriate or at least too expensive (didn't check) to do for all failures.

programs/mango-v4/tests/cases/test_health_compute.rs
```diff
@@ -68,7 +68,7 @@ async fn test_health_compute_tokens() -> Result<(), TransportError> {
     let avg_cu_increase = cu_measurements.windows(2).map(|p| p[1] - p[0]).sum::<u64>()
         / (cu_measurements.len() - 1) as u64;
     println!("average cu increase: {avg_cu_increase}");
-    assert!(avg_cu_increase < 3200);
+    assert!(avg_cu_increase < 3230);
```
ckamm, line 71: should measure the CU needed when a fallback oracle is needed (the logging might be killer)

programs/mango-v4/tests/cases/test_stale_oracles.rs
```diff
@@ -172,3 +176,146 @@ async fn test_stale_oracle_deposit_withdraw() -> Result<(), TransportError> {
...
+    let fallback_oracle = fallback_oracle_kp.pubkey();
+    let admin = TestKeypair::new();
+    let owner = context.users[0].key;
+    let payer = context.users[1].key;
+    let mints = &context.mints[0..3];
+    let payer_token_accounts = &context.users[1].token_accounts[0..3];
+
+    //
+    // SETUP: Create a group, account, register tokens
+    //Foracle_
```
ckamm, line 197: odd comment

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -66,19 +68,22 @@ pub fn new_fixed_order_account_retriever<'a, 'info>(
...
 
     Ok(FixedOrderAccountRetriever {
-        ais: AccountInfoRef::borrow_slice(ais)?,
+        ais: fixed_ais,
         n_banks: active_token_len,
         n_perps: active_perp_len,
         begin_perp: active_token_len * 2,
         begin_serum3: active_token_len * 2 + active_perp_len * 2,
         staleness_slot: Some(Clock::get()?.slot),
+        fallback_oracle_ais,
```
ckamm, line 86: I don't remember why the FixedOrderAccountRetriever has this awkward form - not needing to allocate a vec and just storing the accountinfos would be preferable - but I'd suggest to keep it consistent.

Meaning: keep a single `ais: Vec<T>` and store `fallback_oracle_start: usize` to know what range of accounts are the fallbacks.

Maybe we can change `ais` type to `&[impl KeyedAccountReader]` or something at some point and avoid the allocation. I think the issue was that if I want to hand out plain references, I need to store the `Ref<'a, &[u8]>` somewhere and the whole thing was frustrating.

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -134,7 +134,14 @@ impl<T: KeyedAccountReader> AccountRetriever for FixedOrderAccountRetriever<T> {
             })?;
 
         let oracle_index = self.n_banks + active_token_position_index;
-        let oracle_price = self.oracle_price_bank(oracle_index, bank).with_context(|| {
+        let oracle = &self.ais[oracle_index];
+        let fallback_opt = self
+            .fallback_oracle_ais
+            .iter()
+            .find(|ai| ai.key() == &bank.fallback_oracle);
```
ckamm, line 141: I guess scanning through the list of fallback oracles every time is cheap enough as long as there's only a few...

ckamm (commented):
Some initial comments!

What about the ScanningAccountRetriever? That'd allow it to be used in more instructions.

lib/client/src/health_cache.rs
Lou-Kamades: I haven't given much thought to liquidators yet, what I mostly had in mind was a user who is struggling to place orders. The client could have a bool for including a fallback oracle on the market they are trying to place on for example. Once this is tested I think we could expand how many fallbacks the client provides (I can definitely see use cases where someone would send as many fallbacks as possible)

programs/mango-v4/src/health/account_retriever.rs
Lou-Kamades: Maybe making a HashMap like the ScanningAccountRetirever does would be preferable? Only one scan is guaranteed in that case:
https://github.com/blockworks-foundation/mango-v4/pull/790/commits/91cd42f0443b1f6187aecf67d47d70b7b1a51ad5#diff-0b7a655c1f66e408b1fd89404a260921036deb705eaef311f4ccce13974edd93R373

programs/mango-v4/tests/cases/test_health_compute.rs
Lou-Kamades: Logging is indeed really expensive. I'm seeing a cost of ~16000 CU for each fallback used, with an additional ~11000 if the fallback is also bad

programs/mango-v4/src/health/account_retriever.rs
Lou-Kamades: as far as CU goes, even with 8 tokens it's only max 100 difference (based on `test_health_compute_tokens_fallback_oracles`)

programs/mango-v4/tests/cases/test_health_compute.rs
ckamm: That sounds almost unbelievably high!

programs/mango-v4/src/health/account_retriever.rs
ckamm: Thanks for checking, yeah keep scanning then

programs/mango-v4/src/accounts_ix/token_edit.rs
```diff
@@ -21,4 +21,9 @@ pub struct TokenEdit<'info> {
     ///
     /// CHECK: The oracle can be one of several different account types
     pub oracle: UncheckedAccount<'info>,
+
+    /// The oracle account is optional and only used when reset_stable_price is set.
```
ckamm, line 25: you mean "set_fallback_oracle"

programs/mango-v4/src/health/account_retriever.rs
ckamm: We could also save CU by doing the scan only if the fallback is actually needed (but since CU cost looks ok, this optimization isn't required imo)

programs/mango-v4/tests/cases/test_health_compute.rs
```diff
@@ -142,6 +169,145 @@ async fn test_health_compute_tokens_during_maint_weight_shift() -> Result<(), Tr
...
+    .create(solana)
+    .await;
+
+    let account =
+        create_funded_account(&solana, group, owner, 0, &context.users[1], &[], 1000, 0).await;
+
+    let mut cu_measurements = vec![];
+    for token_account in &context.users[0].token_accounts[..mints.len()] {
+        deposit_cu_datapoint(solana, account, owner, *token_account).await;
+    }
```
ckamm, line 221: might be leftover code?

programs/mango-v4/tests/cases/test_health_compute.rs
```diff
@@ -142,6 +169,145 @@ async fn test_health_compute_tokens_during_maint_weight_shift() -> Result<(), Tr
...
+            i,
+            pair[1],
+            pair[1] - pair[0]
+        );
+    }
+
+    let avg_cu_increase = cu_measurements.windows(2).map(|p| p[1] - p[0]).sum::<u64>()
+        / (cu_measurements.len() - 1) as u64;
+    println!("average cu increase: {avg_cu_increase}");
+    assert!(avg_cu_increase < 16_600);
```
ckamm, line 306: wow, it shouldn't be _that_ expensive, I'll also take a look; let's try to figure out what's eating compute!

programs/mango-v4/tests/cases/test_serum.rs
```diff
@@ -1481,6 +1482,272 @@ async fn test_serum_compute() -> Result<(), TransportError> {
...
+        .serum
+        .list_spot_market(&base_token.mint, &quote_token.mint)
+        .await;
+
+    //
+    // TEST: Register a serum market
+    //
+    let serum_market = send_tx(
+        solana,
+        Serum3RegisterMarketInstruction {
```
ckamm, line 1609: I recommend using common_setup(), check the newer other tests, you probably don't need any unusual settings except for the fallback oracle?

programs/mango-v4/tests/cases/test_health_compute.rs
Lou-Kamades: Maybe I'm being too cautious here but I was trying to isolate the CU of only a stale oracle deposit. I didn't know if creating a token position would overestimate the CU usage of a fallback..

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -196,6 +200,7 @@ impl<T: KeyedAccountReader> AccountRetriever for FixedOrderAccountRetriever<T> {
 pub struct ScannedBanksAndOracles<'a, 'info> {
     banks: Vec<AccountInfoRefMut<'a, 'info>>,
     oracles: Vec<AccountInfoRef<'a, 'info>>,
+    fallback_oracles: HashMap<Pubkey, AccountInfoRef<'a, 'info>>,
```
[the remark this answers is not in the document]
Lou-Kamades: I wrote a simple test that passes fallbacks to a PerpSettlePnl instruction (but doesn't access them). It consumed 86765 vs 79644 without. ~8.8% more CU just to create an unused HashMap indeed seems wasteful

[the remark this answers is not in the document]
Lou-Kamades: Using a vec and scanning in the simple PerpSettlePnl test yielded 81135 with fallbacks vs. 79181 without -- much nicer!

programs/mango-v4/tests/cases/test_health_compute.rs
ckamm: ah, ok!

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -367,25 +379,21 @@ impl<'a, 'info> ScanningAccountRetriever<'a, 'info> {
         let serum3_start = perp_oracles_start + n_perps;
         let n_serum3 = ais[serum3_start..]
             .iter()
-            .filter(|x| serum3_cpi::load_open_orders_ref(x).is_ok())
+            .filter(|x| x.data_len() == std::mem::size_of::<serum_dex::state::OpenOrders>() + 12)
```
ckamm, line 382: - `filter()` isn't the right choice here, maybe `.position(!is_serum_oo)`? 
- might have issues if an oracle randomly has the same size as a serum oo. serum oos always start with the "serum" string, let's check that too

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -377,9 +377,16 @@ impl<'a, 'info> ScanningAccountRetriever<'a, 'info> {
...
             .iter()
-            .filter(|x| x.data_len() == std::mem::size_of::<serum_dex::state::OpenOrders>() + 12)
+            .inspect(|x| {
+                if x.data_len() == std::mem::size_of::<serum_dex::state::OpenOrders>() + 12
+                    && serum3_cpi::has_serum_header(&x.data.borrow())
+                {
+                    n_serum3 += 1;
+                }
+            })
             .count();
```
ckamm, line 390: I still think this is not quite right. What if the user passes "serum oo, oracle, another serum oo" - now n_serum3 would be 2 presumably? That's why I suggest using `slice.iter().position(|x|!is_serum(x)).unwrap_or(slice.len)`.

Or, to keep it similar to what's done for the previous, you could also do `.take_while(|x| is_serum(x)).count()`

programs/mango-v4/src/state/bank.rs
```diff
@@ -947,12 +951,46 @@ impl Bank {
...
+            require_keys_eq!(self.fallback_oracle, *fallback_oracle_acc.key());
+            let fallback_state =
+                oracle::oracle_state_unchecked(fallback_oracle_acc, self.mint_decimals)?;
+            let fallback_ok = fallback_state.check_confidence_and_maybe_staleness(
+                &self.name(),
+                &self.oracle_config,
+                staleness_slot,
+            );
+            fallback_ok.with_context(|| {
+                oracle_log_context(&fallback_state, &self.oracle_config, staleness_slot)
```
ckamm, line 985: could we also log the primary oracle state here? (just both, we'll know that the first is from the primary)

ckamm (approved):

PR #792, opened 2023-11-21
Allow dusting when withdrawing if already zero

brittcyr:
Had a mango account stuck because it had an active token slot used by a token with a stuck oracle.

The UI close button just sends a withdraw, but that does not work when there is already zero in the account.
https://explorer.solana.com/tx/2SgeRVjg1kqhb7wVUcNHoQfPAkH5A5NwQfcHRCxN48iyfuqzCxhrDnrPcDGJJ1w7engSSo99YCpqQxKQkVq9uJGb?cluster=mainnet-beta

I dont know why this wasnt dusted the first time that a withdraw down to zero was done, but subsequent withdraws should dust.
Screen Shot 2023-11-21 at 5 33 46 PM

### DIFF
--- a/programs/mango-v4/src/state/bank.rs
+++ b/programs/mango-v4/src/state/bank.rs
@@ -594,7 +594,7 @@ impl Bank {
         require_gte!(native_amount, 0);
         let native_position = position.native(self);
 
-        if native_position.is_positive() {
+        if native_position.is_positive() || native_position.is_zero() {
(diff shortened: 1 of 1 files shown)


ckamm:
@brittcyr Thanks for the patch, this was exactly right! I didn't see it in time and made my own version https://github.com/blockworks-foundation/mango-v4/pull/793 , unfortunately. :/

PR #797, opened 2023-11-24
WIP: Mc/types

microwavedcola1:


ts/client/src/client.ts
```diff
@@ -321,7 +322,10 @@ export class MangoClient {
 
   public async getGroup(groupPk: PublicKey): Promise<Group> {
     const groupAccount = await this.program.account.group.fetch(groupPk);
-    const group = Group.from(groupPk, groupAccount);
+    const group = Group.from(
+      groupPk,
+      groupAccount as IdlAccounts<MangoV4>['group'],
```
ckamm, line 327: I thought `this.program.account.group.fetch()` and fiends already gave us the right type, so would think that this cast isn't needed.

After all Group.from() demands a specific type for the second argument and there should have been type checking errors if they were incompatible.

ts/client/src/client.ts
```diff
@@ -927,12 +950,12 @@ export class MangoClient {
               v1Len +
                 // This is the padding before tokenConditionalSwaps
                 4,
-            ) as TokenConditionalSwapDto[])
-        : new Array<TokenConditionalSwapDto>();
+            ) as Array<IdlTypes<MangoV4>['TokenConditionalSwap']>)
```
ckamm, line 953: I think just `as`-ing this to the target type is what masks the issue. But note this will be gone anyway once https://github.com/blockworks-foundation/mango-v4/pull/783 is in (which just needs me looking over the ci issues)

microwavedcola1: IMO this doesnt mask the issue, this correctly forces the type on what we get from chain to what we have in the IDL, this is how later when we work these dtos we get compile issues if we dont migrate hasData to isConfigured

I do agree that the DTO stuff in middle is not strictly necessary, but this is required because we want to special handle certain types

ckamm: Oh right, because we're going from a custom Dto struct that manually copies the program-side layout to the actual idl-type. Agree, this is helpful!

microwavedcola1: of course this usage pattern is not enforced and people might forget about using it, I will see if I can enforce a more generic solution

microwavedcola1:
bump: revisit

PR #804, opened 2023-12-05
revert trustless token registration params to conservative defaults

ckamm:
This mostly reverts commit 33bc278a2ca78bf37f291abdc1994e0a3670d57a.


microwavedcola1 (approved):

programs/mango-v4/src/instructions/token_register_trustless.rs
```diff
@@ -87,7 +87,7 @@ pub fn token_register_trustless(
         net_borrows_in_window: 0,
         borrow_weight_scale_start_quote: 5_000_000_000.0, // $5k
```
microwavedcola1, line 88: hmm were these  2 always at this value?

ckamm: yeah - but since borrows are disabled, they don't have an effect

PR #810, opened 2023-12-06
adjust token_register_trustless params

ckamm:
- since there's no borrowing or asset weight, don't put staleness or confidence restrictions on the oracle
- disable the tcs fees, they are not enabled anywhere yet


ckamm:
reviewed by @microwavedcola1 in chat

PR #811, opened 2023-12-06
WIP: Use fallback oracles  in Rust + TS clients

Lou-Kamades:


lib/client/src/context.rs
```diff
@@ -554,6 +598,35 @@ impl MangoGroupContext {
         let new_perp_markets = fetch_perp_markets(rpc, mango_v4::id(), self.group).await?;
         Ok(new_perp_markets.len() > self.perp_markets.len())
     }
+
+    pub async fn fetch_fallbacks_for_stale_oracles(
+        &self,
+        rpc: &RpcClientAsync,
+    ) -> anyhow::Result<Vec<Pubkey>> {
```
ckamm, line 605: This kind of function is what's needed, but I think it should use the AccountFetcher to retrieve the accounts, so users can use caching or websocket feeds or grpc to feed in the actual account data.

Also, there shouldn't be a need to re-fetch the banks, since the context has all the necessary info in `tokens`

I don't see it used yet. If it's using the AccountFetcher, it'd be fine to call it for every tx send.

User config could be
- use dynamic fallback accounts (default): call this function to check what oracles need fallbacks
- pass all fallbacks (your AllAvailable; which has trouble with the 64 accounts-per-tx limit)
- no dynamic fallbacks

Plus a separate config for a fixed list of fallbacks, ideally configured in a user-helpful way (symbol names?). That way even if you use dynamic fallback accounts, you can say "KIN oracle breaks randomly, I'd rather pass it all the time".

lib/client/src/context.rs
```diff
@@ -156,6 +164,8 @@ pub struct MangoGroupContext {
     pub address_lookup_tables: Vec<Pubkey>,
 
     pub compute_estimates: ComputeEstimates,
+
+    pub fallback_oracle_context: FallbackOracleContext,
```
ckamm, line 168: I'm doubtful of putting this user-side configuration into the MangoGroupContext - its intention is to cache the on chain state that's needed frequently and changes rarely.

lib/client/src/context.rs
```diff
@@ -554,6 +598,35 @@ impl MangoGroupContext {
...
+        let now_slot = rpc.get_slot().await?;
+        let mut fallbacks = vec![];
+        for acc in oracle_accounts {
+            let bank = banks_by_oracle.get(acc.key()).unwrap();
+            let state = oracle_state_unchecked(&acc, bank.mint_decimals)?;
+            if (state
+                .check_confidence(bank.name(), &bank.oracle_config)
+                .is_err()
+                || state
+                    .check_staleness(bank.name(), &bank.oracle_config, now_slot)
```
ckamm, line 621: could use `check_confidence_and_maybe_staleness()`

ckamm (commented):
only covered rust part for now

PR #813, opened 2023-12-08
Add an Orca oracle type

Lou-Kamades:


programs/mango-v4/src/state/orca_cpi.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    pub reward_infos: [WhirlpoolRewardInfo; NUM_REWARDS], // 384
+}
+
+// Number of rewards supported by Whirlpools
+pub const NUM_REWARDS: usize = 3;
+
+impl Whirlpool {
+    pub const LEN: usize = 8 + 261 + 384;
+    pub fn seeds(&self) -> [&[u8]; 6] {
```
ckamm, line 57: not needed?

ckamm (commented):
I'll need another more detailed review pass...

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -465,12 +467,22 @@ impl<'a, 'info> AccountRetriever for ScanningAccountRetriever<'a, 'info> {
     }
 }
 
+/// Contains options for a fallback oracle and the Pyth USDC oracle if they were provided in the remaining accounts.
+/// The Pyth USDC oracle is only used if the fallback oracle type is a CLMM
+pub type FallbackOracleInfos<'a, T> = (Option<&'a T>, Option<&'a T>);
```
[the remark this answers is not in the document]
ckamm: New suggestion:
- we support only USDC and SOL quote types
- let's not scan for these oracles every time we read an oracle; can we scan for both once when making the account retriever?
- reading a orca oracle will check which side is usdc or sol and auto-flip based on that

ckamm:
I'd like to see an integration test that uses an orca oracle. It's reasonably easy to do now that ProgramTestContext has set_account(). (our solana.rs has a new set_account() for anchor-style accounts, added fairly recently in dev - but can also just set accounts to some bytes you read from a file that way)

programs/mango-v4/src/instructions/token_deposit.rs
```diff
@@ -92,6 +92,7 @@ impl<'a, 'info> DepositCommon<'a, 'info> {
         // to deposit to close borrows or do other fixes even if the oracle is bad.
         let unsafe_oracle_state = oracle_state_unchecked(
             &AccountInfoRef::borrow(self.oracle.as_ref())?,
+            EMPTY_KEYED_READER_OPT,
```
ckamm, line 95: fwiw, I found `EMPTY_KEYED_READER_OPT` more confusing then `None`. It could be `NO_USDC_PYTH_FEED` if you intend this for this particular function only? But it doesn't feel quite right; make a new function? (that can potentially forward to a private function with an optional argument)

Key issue is that oracle_state_unchecked() now has two variants: the one for primary oracles that doesn't allow orca, and the one for fallbacks that does but needs extra info.

Also all instructions that manually read oracles outside of AccountRetriever will not be able to use fallbacks. that includes token_deposit and token_withdraw (at least for some uses)

programs/mango-v4/src/state/oracle.rs
```diff
@@ -317,9 +356,53 @@ pub fn oracle_state_unchecked(
                 oracle_type: OracleType::SwitchboardV1,
             }
         }
+        OracleType::OrcaCLMM => {
+            let usd_state = usdc_state_unchecked(usd_feed_opt)?;
+
+            let whirlpool = Whirlpool::try_deserialize(&mut &data[..]).unwrap();
```
ckamm, line 362: I'm curious about measuring the cost of this. It uses anchor's borsh deserialization to fill a bunch of fields we don't even read. The data looks to be in a fixed layout, we could probably zero-copy access it instead to save a bunch of deserialization overhead

programs/mango-v4/src/state/oracle.rs
```diff
@@ -317,9 +356,53 @@ pub fn oracle_state_unchecked(
...
+
+            let sqrt_price = U64F64::from_bits(whirlpool.sqrt_price);
+
+            let clmm_price = if whirlpool.token_mint_a == usdc_mint_mainnet::ID {
+                // inverted
+                let decimals = QUOTE_DECIMALS - (base_decimals as i8); // tokenMintB - tokenMintA
+                let decimal_adj = power_of_ten(decimals);
+                let inv_price = I80F48::from_num(sqrt_price * sqrt_price) * decimal_adj;
```
ckamm, line 370: On first reading this, I was worried that there might not be enough precision to do the squaring in U64F64-space in particular if we drop the last 16 bits of precision afterwards. But it looks ok: after all we assume the price in the end has good precision in I80F48.

But: did you check the decimals logic? My expectation would be that sqrt_price is already in native/native units and not in ui units, so all the decimal shifting should be unnecessary. (verify by looking at the SOL/USDC sqrt_price)

programs/mango-v4/src/state/oracle.rs
ckamm: (since it's not aligned, we can't make a struct, but we could just manually take bytes N..M and intrepret them as the sqrt_price)

programs/mango-v4/src/state/oracle.rs
```diff
@@ -317,9 +356,53 @@ pub fn oracle_state_unchecked(
...
+                let decimal_adj = power_of_ten(decimals);
+                Ok(I80F48::from_num(sqrt_price * sqrt_price) * decimal_adj)
+            } else {
+                Err(MangoError::InvalidCLMMOracle)
+            }?;
+
+            let price = clmm_price * usd_state.price;
+            OracleState {
+                price,
+                last_update_slot: u64::MAX, // REVIEW: should we inherit pyth feed confidence metrics?
```
ckamm, line 383: I think this should be `min(pyth.last_update_slot, orca.last_update_slot)` - and if we don't have one for orca, then just pyth

Lou-Kamades: sadly, orca does not store last update slot

programs/mango-v4/src/state/oracle.rs
Lou-Kamades: ah yes good catch! The price is already in native/native

programs/mango-v4/src/state/oracle.rs
Lou-Kamades: Manually taking the bytes saves roughly 1000 CU

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -38,14 +38,16 @@ pub fn account_buyback_fees_with_mngo(
     let now_ts = clock.unix_timestamp.try_into().unwrap();
     let slot = clock.slot;
 
+    let mngo_oracle_ref = &AccountInfoRef::borrow(&ctx.accounts.mngo_oracle.as_ref())?;
     let mngo_oracle_price = mngo_bank.oracle_price(
-        &AccountInfoRef::borrow(&ctx.accounts.mngo_oracle.as_ref())?,
+        &OracleAccountInfos::from_reader(mngo_oracle_ref),
```
ckamm, line 43: This looks a bit annoying to me, because of the awkwardness of having to make an AccountInfoRef (which just stores a collection of references) and then store a _reference_ to that in OracleAccountInfos (now a collection of references to things that have references).

I suspect that `&AccountInfoRef` should also implement KeyedAccountReader and then a lot of issues would go away, because then we could probably make a `OracleAccountInfos::from_ai()` helper that returns a `OracleAccountInfos<&AccountInfoRef>` without needing to make a named temporary solely for ownership reasons.

I'd be up for making that change, doesn't need to be part of this PR.

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
```diff
@@ -38,14 +38,16 @@ pub fn account_buyback_fees_with_mngo(
     let now_ts = clock.unix_timestamp.try_into().unwrap();
     let slot = clock.slot;
 
+    let mngo_oracle_ref = &AccountInfoRef::borrow(&ctx.accounts.mngo_oracle.as_ref())?;
```
ckamm, line 41: aside: making these Refs like this can be dangerous because if the rust compiler doesn't drop() them asap, a future attempted mutable borrow could fail _at runtime_; it won't be an issue because no-one borrows oracles mutably, but it's something to be aware of. Previously the temporary values were guaranteed to be dropped at the end of the line.

programs/mango-v4/src/instructions/perp_liq_negative_pnl_or_bankruptcy.rs
```diff
@@ -519,25 +520,33 @@ mod tests {
...
-                .oracle_price(&AccountInfoRef::borrow(&settle_oracle_ai).unwrap(), None)
-                .unwrap();
-            let perp_price = setup
-                .perp_market
-                .data()
-                .oracle_price(&AccountInfoRef::borrow(&perp_oracle_ai).unwrap(), None)
-                .unwrap();
+            let insurance_price = {
+                let insurance_oracle_ai = setup.insurance_oracle.as_account_info();
+                let insurance_oracle_ref = &AccountInfoRef::borrow(&insurance_oracle_ai)?;
```
ckamm, line 525: while you're touching this code, prefer as_ref() over as_account_info() because the latter clone()s the AccountInfo, which isn't needed here

programs/mango-v4/src/state/bank.rs
```diff
@@ -1065,14 +1064,13 @@ impl Bank {
         self.interest_curve_scaling = (self.interest_curve_scaling * adjustment).max(1.0)
     }
 
-    pub fn oracle_price(
+    pub fn oracle_price<T: KeyedAccountReader>(
         &self,
-        oracle_acc: &impl KeyedAccountReader,
+        oracle_acc_infos: &OracleAccountInfos<T>,
         staleness_slot: Option<u64>,
     ) -> Result<I80F48> {
```
ckamm, line 1071: honest question: now that oracle_price and oracle_price_with_fallback have the same signature, do we even still need both? Maybe it's enough to say "the fallback is checked, if `oracle_acc_infos.fallback.is_some()`"?

programs/mango-v4/src/state/oracle.rs
```diff
@@ -357,50 +412,55 @@ pub fn oracle_state_unchecked(
...
+                    && whirlpool.token_mint_b != usdc_mint_mainnet::ID);
+            let quote_state = if inverted {
+                quote_state_unchecked(acc_infos, &whirlpool.token_mint_a)?
+            } else {
+                quote_state_unchecked(acc_infos, &whirlpool.token_mint_b)?
+            };
+
+            let clmm_price = if inverted {
+                let inverted_price = I80F48::from_num(sqrt_price * sqrt_price);
+                I80F48::ONE.checked_div(inverted_price).unwrap()
```
ckamm, line 429: 1. no need to use checked_div, checked math is enabled for all I80F48 ops
2. we likely want to use f64 division anyway because it's way cheaper (2000 CU vs 200 CU)

For the latter, the cheapest path would be to convert sqrt_price to f64 very early (costs 100 CU), then do the squaring (200 CU cheaper as f64), do the division (1800 CU cheaper as f64) and then convert back to I80F48 (costs 200 CU). It's sad that this is only cheaper when the division is needed, so need two different codepaths?

programs/mango-v4/src/state/orca_cpi.rs
```diff
@@ -1,72 +1,47 @@
...
+    require!(
+        data.len() == ORCA_WHIRLPOOL_LEN,
+        MangoError::InvalidCLMMOracle
+    );
+    require!(
+        acc_info.owner() == &orca_mainnet_whirlpool::ID,
+        MangoError::InvalidCLMMOracle
+    );
+
+    let sqrt_price = u128::try_from_slice(&data[65..81])?;
```
ckamm, line 36: imo, be specific and use from_le_bytes()

programs/mango-v4/tests/cases/test_stale_oracles.rs
```diff
@@ -316,3 +319,196 @@ async fn test_fallback_oracle_withdraw() -> Result<(), TransportError> {
...
+    // setup pyth and clmm accounts
+    for fixture in fixtures {
+        let filename = format!("resources/test/{}.bin", fixture.0);
+        let data = read_file(find_file(&filename).unwrap());
+        let mut account =
+            AccountSharedData::new(u64::MAX, data.len(), &Pubkey::from_str(fixture.1).unwrap());
+        account.set_data(data);
+        let mut program_test_context = solana.context.borrow_mut();
+        program_test_context.set_account(&Pubkey::from_str(fixture.0).unwrap(), &account);
+    }
```
ckamm, line 357: :+1:

programs/mango-v4/src/instructions/account_buyback_fees_with_mngo.rs
ckamm: I played with it a bit and don't think there's a straightforward way to make this nicer. The key problem always is that AccountInfos have a RefCell and someone needs to hold on to the Ref instance if we want to have long-lived references to account data.

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -227,12 +259,13 @@ impl<'a, 'info> ScannedBanksAndOracles<'a, 'info> {
     ) -> Result<(&mut Bank, I80F48, Option<(&mut Bank, I80F48)>)> {
         if token_index1 == token_index2 {
             let index = self.bank_index(token_index1)?;
+            let price = {
+                let bank = self.banks[index].load_fully_unchecked::<Bank>()?;
+                let oracle_acc_infos = self.create_oracle_infos(index, Some(&bank.fallback_oracle));
+                bank.oracle_price_with_fallback(&oracle_acc_infos, self.staleness_slot)?
+            };
```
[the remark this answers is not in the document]
Lou-Kamades: Since these function take in `&mut self` (where self is some sort of AccountRetriever) `let bank = self.banks[index].load_mut_fully_unchecked::<Bank>()?;` is a mutable borrow. At the same time`self.create_oracle_infos(index, Some(&bank.fallback_oracle));` is immutable.

I decided to drop the immutable borrow before the mutable one to make the borrow checker happy but I'm open to alternatives. This was the least-invasive way I could think of

programs/mango-v4/src/instructions/perp_liq_negative_pnl_or_bankruptcy.rs
Lou-Kamades: These are `TestAccount`s but I ensured that any instruction uses .as_ref()

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -225,12 +258,13 @@ impl<'a, 'info> ScannedBanksAndOracles<'a, 'info> {
     ) -> Result<(&mut Bank, I80F48, Option<(&mut Bank, I80F48)>)> {
         if token_index1 == token_index2 {
             let index = self.bank_index(token_index1)?;
+            let price = {
+                let bank = self.banks[index].load_fully_unchecked::<Bank>()?;
+                let oracle_acc_infos = self.create_oracle_infos(index, &bank.fallback_oracle);
```
ckamm, line 263: this function takes the oracle_index as the first argument, not the bank index, so this is likely a bug. It might be helpful to have a `bank_oracle_infos()` fn where you pass in the bank index, it reads that to get the fallback, computes the oracle index etc?

programs/mango-v4/src/health/account_retriever.rs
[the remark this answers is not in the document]
ckamm: Ah yes, unfortunate but necessary. Agree this is the simplest way to do it when faced with mutable borrows like this.

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -267,14 +302,33 @@ impl<'a, 'info> ScannedBanksAndOracles<'a, 'info> {
         let index = self.bank_index(token_index)?;
         // The account was already loaded successfully during construction
         let bank = self.banks[index].load_fully_unchecked::<Bank>()?;
-        let oracle = &self.oracles[index];
-        let fallback_oracle_opt =
-            fetch_fallback_oracle(&self.fallback_oracles, &bank.fallback_oracle);
-        let price =
-            bank.oracle_price_with_fallback(oracle, fallback_oracle_opt, self.staleness_slot)?;
+        let oracle_acc_infos = self.create_oracle_infos(index, &bank.fallback_oracle);
```
ckamm, line 305: also bank and oracle index confusion

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -241,21 +275,22 @@ impl<'a, 'info> ScannedBanksAndOracles<'a, 'info> {
             (index2, index1, true)
         };
 
+        let (price1, price2) = {
+            let (first_bank_part, second_bank_part) = self.banks.split_at(first + 1);
+            let bank1 = first_bank_part[first].load_fully_unchecked::<Bank>()?;
+            let bank2 = second_bank_part[second - (first + 1)].load_fully_unchecked::<Bank>()?;
+            let oracle_infos_1 = self.create_oracle_infos(first, &bank1.fallback_oracle);
+            let oracle_infos_2 = self.create_oracle_infos(second, &bank2.fallback_oracle);
```
ckamm, line 283: indexes

programs/mango-v4/src/state/oracle.rs
```diff
@@ -425,9 +425,11 @@ fn oracle_state_unchecked_inner<T: KeyedAccountReader>(
             };
 
             let clmm_price = if inverted {
-                let inverted_price = I80F48::from_num(sqrt_price * sqrt_price);
-                I80F48::ONE.checked_div(inverted_price).unwrap()
+                let sqrt_price = f64::from_fixed(U64F64::from_bits(whirlpool.sqrt_price));
```
ckamm, line 428: minor, for consistency: we've typically used `.to_num::<f64>()` for this conversion - prefer to use that here if that's allowed

programs/mango-v4/tests/cases/test_health_compute.rs
```diff
@@ -344,7 +344,7 @@ async fn test_health_compute_tokens_fallback_oracles() -> Result<(), TransportEr
 #[tokio::test]
 async fn test_health_compute_serum() -> Result<(), TransportError> {
     let mut test_builder = TestContextBuilder::new();
-    test_builder.test().set_compute_max_units(130_500);
+    test_builder.test().set_compute_max_units(135_000);
```
ckamm, line 347: huh, how did this patch raise the CU cost this much?

ckamm:
if the index thing is a real issue, then this code is undertested (and has been before); would much appreciate adding a test that would at least detect essential issues like this one

programs/mango-v4/src/health/account_retriever.rs
Lou-Kamades: For a ScanningAccountRetriever, bank index == oracle index (since banks and oracles are stored in separate vectors)

This is not the case in a FixedAccountRetriever however and `create_oracle_infos` uses `self.ais[index]` to pick the oralce account  vs `self.oracles[index]` (since all account infos are stored together)

ckamm: Right, got it, not an issue.

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -241,21 +275,22 @@ impl<'a, 'info> ScannedBanksAndOracles<'a, 'info> {
             (index2, index1, true)
         };
 
+        let (price1, price2) = {
+            let (first_bank_part, second_bank_part) = self.banks.split_at(first + 1);
+            let bank1 = first_bank_part[first].load_fully_unchecked::<Bank>()?;
+            let bank2 = second_bank_part[second - (first + 1)].load_fully_unchecked::<Bank>()?;
```
ckamm, line 281: Reading this again, I think it can be simplified to
```
            let bank1 = self.banks[first].load_fully_unchecked::<Bank>()?;
            let bank2 = self.banks[second].load_fully_unchecked::<Bank>()?;
```
because with immutable borrows we don't need to split the banks Vec into two parts.

ckamm (approved):
(please squash-merge)

PR #821, opened 2023-12-19
ts-client optimitics ui confirmations - draft

abrzezinski94:


ts/client/src/client.ts
```diff
@@ -2726,7 +2801,7 @@ export class MangoClient {
       group,
       [hrix1, ...ixs, hrix2],
       {
-        prioritizationFee: true,
```
abrzezinski94, line 2729: Made fixes here this was never working

PR #836, opened 2024-01-04
Openbook V2 Integration

riordanp:
ckamm's pending:
- [x] is there a program test for obv2_settle?
- [x] would be good to have a test for reading old account data with borsh and zerocopy? (moved to #878 )
- [x] tests don't pass right now
- [x] one obv2 program test still commented out until program fix and .so update
- [x] liquidator needs to be updated to force cancel obv2 orders


programs/mango-v4/src/state/mango_account_components.rs
```diff
@@ -192,6 +192,89 @@ impl Default for Serum3Orders {
...
+    /// Tracks the amount of deposits that flowed into the serum open orders account.
+    ///
+    /// The bank still considers these amounts user deposits (see deposits_in_serum)
+    /// and they need to be deducted from there when they flow back into the bank
+    /// as real tokens.
+    pub base_deposits_reserved: u64,
+    pub quote_deposits_reserved: u64,
+
+    #[derivative(Debug = "ignore")]
+    pub reserved: [u8; 32],
```
ckamm, line 240: maybe want to reserve more bytes to future-proof? it's likely that openbook v2 will evolve quite a bit still

programs/mango-v4/tests/cases/test_alt.rs
```diff
@@ -30,7 +30,7 @@ async fn test_alt() -> Result<(), TransportError> {
 
     let address_lookup_table = solana.create_address_lookup_table(payer, payer).await;
 
-    send_tx(
+    mango_client::send_tx(
```
ckamm, line 33: send_tx is used alot in all the tests, is this needed?

programs/mango-v4/src/health/cache.rs
```diff
@@ -287,6 +288,33 @@ impl Serum3Info {
...
+
+        Self {
+            reserved_base,
+            reserved_quote,
+            reserved_base_as_quote_lowest_ask,
+            reserved_quote_as_base_highest_bid,
+            base_info_index,
+            quote_info_index,
+            market_index: open_orders.market_index,
+            has_zero_funds: open_orders_account.position.is_empty(),
```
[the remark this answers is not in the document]
riordanp: this should check all the equivalent values to the above serum3 check https://github.com/openbook-dex/openbook-v2/blob/master/programs/openbook-v2/src/state/open_orders_account.rs#L393

maybe we don't care about fees etc here and only want to know about open orders or filled and unprocessed orders, in which case the has_open_orders helper on the position should be used instead

programs/mango-v4/src/health/cache.rs
```diff
@@ -966,13 +1038,13 @@ impl HealthCache {
...
-        self.has_serum3_open_orders_funds() || self.has_perp_open_orders()
+        self.has_spot_open_orders_funds() || self.has_perp_open_orders()
     }
 
     pub fn require_after_phase1_liquidation(&self) -> Result<()> {
         require!(
-            !self.has_serum3_open_orders_funds(),
-            MangoError::HasOpenOrUnsettledSerum3Orders
+            !self.has_spot_open_orders_funds(),
+            MangoError::HasOpenOrUnsettledSerum3Orders // todo-pan: change error?
```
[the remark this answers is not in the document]
riordanp: yeah it's a bit messy, through most of the pr I opted to duplicate code to preserve backwards compatibility but it breaks down here, not sure what the best way forward is

ckamm (commented):
`MangoAccount::expected_health_accounts()` needs an update for obv2

programs/mango-v4/src/health/cache.rs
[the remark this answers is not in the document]
ckamm: I see, then this looks correct. `has_zero_funds` is all about liquidation: if it's false, then force_cancel_and_settle must happen before further liquidations are allowed

programs/mango-v4/src/health/cache.rs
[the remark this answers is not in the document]
ckamm: Keep the number, but I think it's fine to change the name/string

programs/mango-v4/src/health/account_retriever.rs
```diff
@@ -429,7 +461,14 @@ impl<'a, 'info> ScanningAccountRetriever<'a, 'info> {
                     && serum3_cpi::has_serum_header(&x.data.borrow())
             })
             .count();
-        let fallback_oracles_start = serum3_start + n_serum3;
+        let openbook_v2_start = serum3_start + n_serum3;
+        let n_openbook_v2 = ais[openbook_v2_start..]
+            .iter()
+            .take_while(|x| {
+                x.data_len() == std::mem::size_of::<openbook_v2::state::OpenOrdersAccount>() + 8
```
ckamm, line 468: just using the length isn't sufficiently safe against accidents - maybe a fallback oracle happens to have the same data length. Should at least check the discriminator. We would love to check the owner, but if it's like obv1, then there could be different programs?

programs/mango-v4/src/instructions/openbook_v2_cancel_all_orders.rs
```diff
@@ -0,0 +1,78 @@
...
+    //     mango_account: ctx.accounts.account.key(),
+    //     market_index: serum_market.market_index,
+    //     base_token_index: serum_market.base_token_index,
+    //     quote_token_index: serum_market.quote_token_index,
+    //     base_total: after_oo.native_base_total(),
+    //     base_free: after_oo.native_base_free(),
+    //     quote_total: after_oo.native_quote_total(),
+    //     quote_free: after_oo.native_quote_free(),
+    //     referrer_rebates_accrued: after_oo.native_rebates(),
+    // });
```
ckamm, line 56: todo: logging (also in other instructions)

programs/mango-v4/src/instructions/openbook_v2_cancel_all_orders.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    Ok(())
+}
+
+fn cpi_cancel_all_orders(ctx: &OpenbookV2CancelOrder, seeds: &[&[&[u8]]], limit: u8) -> Result<()> {
+    let cpi_accounts = CancelOrder {
+        signer: ctx.account.to_account_info(),
```
ckamm, line 63: using the mango account as the signer instead of the group has some advantages (safety!, easier filtering for user-owned oos) and disadvantages (account must be passed, not easy to filter for all mango-related oo accounts).

programs/mango-v4/src/instructions/openbook_v2_cancel_all_orders.rs
```diff
@@ -0,0 +1,78 @@
...
+    };
+
+    let cpi_ctx = CpiContext::new_with_signer(
+        ctx.openbook_v2_program.to_account_info(),
+        cpi_accounts,
+        seeds,
+    );
+
+    // todo-pan: maybe allow passing side for cu opt
+    openbook_v2::cpi::cancel_all_orders(cpi_ctx, None, limit)
```
ckamm, line 77: yeah, it'd be nice to expose cancelling all on a side - but not important, I think

programs/mango-v4/src/instructions/openbook_v2_close_open_orders.rs
```diff
@@ -0,0 +1,71 @@
...
+    );
+
+    //
+    // close OO
+    //
+    let account_seeds = mango_account_seeds!(account.fixed);
+    cpi_close_open_orders(ctx.accounts, &[account_seeds])?;
+
+    // Reduce the in_use_count on the token positions - they no longer need to be forced open.
+    // We cannot immediately dust tiny positions because we don't have the banks.
```
ckamm, line 40: this points towards us possibly passing the banks here now that we have the opportunity

programs/mango-v4/src/instructions/openbook_v2_close_open_orders.rs
```diff
@@ -0,0 +1,71 @@
...
+    let (base_position, _) = account.token_position_mut(openbook_market.base_token_index)?;
+    base_position.decrement_in_use();
+    let (quote_position, _) = account.token_position_mut(openbook_market.quote_token_index)?;
+    quote_position.decrement_in_use();
+
+    // Deactivate the serum open orders account itself
```
ckamm, line 46: not really serum

programs/mango-v4/src/instructions/openbook_v2_close_open_orders.rs
```diff
@@ -0,0 +1,71 @@
...
+    account.deactivate_openbook_v2_orders(openbook_market.market_index)?;
+
+    Ok(())
+}
+
+fn cpi_close_open_orders(ctx: &OpenbookV2CloseOpenOrders, seeds: &[&[&[u8]]]) -> Result<()> {
+    // todo-pan: when do we clean up the indexer? new ix? can we tell here if nothing else is using it?
```
ckamm, line 53: open question

programs/mango-v4/src/accounts_ix/openbook_v2_place_order.rs
```diff
@@ -50,27 +52,23 @@ pub struct OpenbookV2PlaceOrder<'info> {
...
+    pub market_vault: Box<Account<'info, TokenAccount>>,
 
     /// CHECK: Validated by the openbook_v2 cpi call
     pub market_vault_signer: UncheckedAccount<'info>,
 
     /// The bank that pays for the order, if necessary
     // token_index and payer_bank.vault == payer_vault is validated inline at #3
     #[account(mut, has_one = group)]
-    pub payer_bank: AccountLoader<'info, Bank>,
+    pub bank: AccountLoader<'info, Bank>,
```
ckamm, line 64: There's been a bunch of pain around not having a writable payer_bank and receiver_bank in serum3_place_order. Strongly suggest passing payer_bank (mut) and receiver_bank (mut) explicitly. (and then it makes sense to keep the payer_ prefix again)

The oracles we should start reading from the health accounts instead, so those aren't essential and will start being ignored in the future.

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -0,0 +1,565 @@
...
+    let before_oo_free_slots;
+    let before_had_bids;
+    let before_had_asks;
+    let before_oo = {
+        let open_orders = ctx.accounts.open_orders.load()?;
+        before_oo_free_slots = MAX_OPEN_ORDERS
+            - open_orders
+                .all_orders_in_use()
+                .collect::<Vec<&OpenOrder>>()
+                .len();
```
ckamm, line 129: skip the .collect() and just use .count() on the iterator for a cu and memory reduction (same below)

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -0,0 +1,565 @@
...
+        before_had_bids = open_orders.position.bids_base_lots != 0;
+        before_had_asks = open_orders.position.asks_base_lots != 0;
```
ckamm, line 131: In serum3, why did I do that complicated check with free_slot_bits and is_bid_bits when I could just have checked reserved > 0? Now I feel stupid. :P

programs/mango-v4/src/lib.rs
```diff
@@ -1624,88 +1686,119 @@ pub mod mango_v4 {
...
-        max_base_qty: u64,
-        max_native_quote_qty_including_fees: u64,
-        self_trade_behavior: u8, // openbook_v2::state::SelfTradeBehavior
-        order_type: u8,          // openbook_v2::state::PlaceOrderType
+        side: openbook_v2::state::Side,
+        price_lots: i64,
+        max_base_lots: i64,
+        max_quote_lots_including_fees: i64,
         client_order_id: u64,
+        order_type: openbook_v2::state::PlaceOrderType,
```
ckamm, line 1733: need to double check that anchor can deal with these foreign enums when it generates the idl

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -0,0 +1,565 @@
...
+
+    //
+    // CPI to place order
+    //
+    let account_seeds = mango_account_seeds!(account.fixed);
+    cpi_place_order(
+        ctx.accounts,
+        &[account_seeds],
+        &order,
+        limit.try_into().unwrap(),
```
ckamm, line 168: why pass an u16 limit when anchor only allows u8?

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -0,0 +1,565 @@
...
+    )?;
+
+    //
+    // After-order tracking
+    //
+    let open_orders = ctx.accounts.open_orders.load()?;
+    let after_oo_free_slots;
+    let after_oo = {
```
ckamm, line 176: these braces seem useless

programs/mango-v4/src/accounts_ix/openbook_v2_place_order.rs
```diff
@@ -50,27 +49,23 @@ pub struct OpenbookV2PlaceOrder<'info> {
...
     /// The bank vault that pays for the order, if necessary
     #[account(mut)]
-    pub payer_vault: Box<Account<'info, TokenAccount>>,
+    pub vault: Box<Account<'info, TokenAccount>>,
 
     /// CHECK: The oracle can be one of several different account types
-    #[account(address = payer_bank.load()?.oracle)]
-    pub payer_oracle: UncheckedAccount<'info>,
+    #[account(address = bank.load()?.oracle)]
+    pub oracle: UncheckedAccount<'info>,
```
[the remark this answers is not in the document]
riordanp: can I move them to remaining_accounts now and use bank_and_oracle from FixedOrderAccountRetriever to fetch them?

programs/mango-v4/src/health/account_retriever.rs
riordanp: added discriminator check so it's at least the same as v1, yeah possible there could be more than one deployment, could have a list of whitelisted program ids maybe

programs/mango-v4/src/accounts_ix/openbook_v2_place_order.rs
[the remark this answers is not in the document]
ckamm: Yes, that'd be the future-proof approach

programs/mango-v4/src/instructions/openbook_v2_register_market.rs
```diff
@@ -13,24 +13,49 @@ pub fn openbook_v2_register_market(
...
+        // C tier tokens (no borrows, no asset weight) allow wider bands if the quote token has
+        // no deposit limits
+        let base_c_tier =
+            base_bank.are_borrows_reduce_only() && base_bank.maint_asset_weight.is_zero();
+        let quote_has_no_deposit_limit = quote_bank.deposit_weight_scale_start_quote == f64::MAX
+            && quote_bank.deposit_limit == 0;
+        if base_c_tier && quote_has_no_deposit_limit {
+            require_eq!(oracle_price_band, 19.0);
+        } else {
+            require_eq!(oracle_price_band, 0.5);
```
ckamm, line 53: this was updated to 1

programs/mango-v4/src/state/mango_account.rs
```diff
@@ -311,6 +311,7 @@ impl MangoAccount {
             perp_oo_count,
             token_conditional_swap_count,
         ) + (BORSH_VEC_SIZE_BYTES + size_of::<OpenbookV2Orders>() * usize::from(openbook_v2_count))
+            + BORSH_VEC_PADDING_BYTES
```
ckamm, line 314: I don't think this is right, the reserved bytes are not a vec and need no padding

ckamm:
Open
- is there a program test for obv2_settle?
- is there a test for reading old account data with borsh and zerocopy?

programs/mango-v4/src/state/mango_account.rs
```diff
@@ -1857,19 +2078,40 @@ mod tests {
...
+            fixed: bytemuck::from_bytes_mut(fixed),
+            dynamic,
+        };
+        account.write_token_conditional_swap_length();
+        account.write_openbook_v2_length();
+        let mut account = MangoAccountValue::from_bytes(&bytes).unwrap();
+        // Initialize the openbook orders with defaults as they would be in the program
+        for i in 0..header.openbook_v2_count() {
+            *account.openbook_v2_orders_mut_by_raw_index(i) = OpenbookV2Orders::default();
+        }
```
ckamm, line 2112: now that you've changed `MangoAccount` most of the modifications to this function can go away

programs/mango-v4/src/state/mango_account.rs
```diff
@@ -1900,10 +2142,26 @@ mod tests {
...
+        let account_bytes = {
+            let mut b = account_bytes_without_tcs_and_obv2_and_reserved.clone();
+            // tcs adds 4 bytes of padding and 4 bytes of Vec size
+            b.extend([0u8; 8]);
+            // openbook v2 probably adds a vec too
+            b.extend([0u8; 8]);
+            // plus 56 bytes of reserved space at the end
+            b.extend([0u8; 56]);
+            b
+        };
```
ckamm, line 2156: the try_to_vec() output should now already have the right format

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -0,0 +1,565 @@
...
+                .collect::<Vec<&OpenOrder>>()
+                .len();
+        OpenOrdersSlim::from_oo_v2(&open_orders)
+    };
+    let oo_difference = OODifference::new(&before_oo, &after_oo);
+
+    //
+    // Track the highest bid and lowest ask, to be able to evaluate worst-case health even
+    // when they cross the oracle
+    //
```
ckamm, line 189: Is updating a min/max approximation even needed for openbook v2? Openbook v2 tracks the `locked_price` for each OpenOrder, which should be good enough. And since there can be only 24 open orders at a time, looping over them wouldn't be too expensive.

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -0,0 +1,565 @@
...
+            health_cache,
+            serum_account,
+            open_orders,
+        )?;
+    }
+
+    Ok(())
+}
+
+fn update_bank_potential_tokens_v2(
```
ckamm, line 492: since this is a private function, the _v2 suffix confused me. probably you did this for consistency with `apply_settle_changes_v2` which is public? Maybe we could always refer to `openbook_v2_place_order::apply_settle_changes` explicitly and not `use ..:::*;` in lib.rs? hmm

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
ckamm: That means we might be able to update max_quote and max_base to specific values based on the open orders state, instead of overestimating them!

programs/mango-v4/src/health/cache.rs
```diff
@@ -230,14 +231,14 @@ impl TokenInfo {
...
-/// Information about reserved funds on Serum3 open orders accounts.
+/// Information about reserved funds on Serum3 and Openbook V2 open orders accounts.
 ///
 /// Note that all "free" funds on open orders accounts are added directly
 /// to the token info. This is only about dealing with the reserved funds
 /// that might end up as base OR quote tokens, depending on whether the
 /// open orders execute on not.
 #[derive(Clone, Debug)]
-pub struct Serum3Info {
+pub struct SpotInfo {
```
ckamm, line 241: Significant issue: `SpotInfo::market_index`: This will be 0 both for the 0th serum3 market and the 0th openbook market, causing issues for all functions that try to look up a SpotInfo based on a serum3 or openbook market index.

Propably store `serum3_index : Option<Serum3MarketIndex, obv2_index: Option<OpenBookV2MarketIndex>`? This would have been seen at compile time if these indexes weren't just u16...

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -0,0 +1,565 @@
...
+
+    // Check the receiver's reduce only flag.
+    //
+    // Note that all orders on the book executing can still cause a net deposit. That's because
+    // the total openbook_v2 potential amount assumes all reserved amounts convert at the current
+    // oracle price.
+    if receiver_bank_reduce_only {
+        let balance = health_cache.token_info(receiver_token_index)?.balance_spot;
+        let potential =
+            health_cache.total_spot_potential(HealthType::Maint, receiver_token_index)?; // todo-pan: split potential into serum and openbook
```
ckamm, line 376: hmm, shouldn't it stay combined? we don't want to cause deposits even if all orders over all relevant markets execute

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -0,0 +1,565 @@
...
+
+    // Provide a readable error message in case the vault doesn't have enough tokens
+    let base_lot_size;
+    let quote_lot_size;
+    {
+        let openbook_market_external = ctx.accounts.openbook_v2_market_external.load()?;
+        base_lot_size = openbook_market_external.base_lot_size;
+        quote_lot_size = openbook_market_external.quote_lot_size;
+
+        // todo-pan: why i64? hope the cast doesnt fuck anything up
```
ckamm, line 143: yeah, I saw `require_gte!(..., 0)` in the obv2 code, but agree that some explicit
```
let max_base_lots: u64 = order.max_base_lots.try_into().unwrap();
```
etc would me feel better to be safe (-1 as u64 would really mess things up)

ckamm:
There's heaps of code duplication that I didn't go through in detail, makes me a bit nervous

programs/mango-v4/src/accounts_ix/openbook_v2_place_order.rs
```diff
@@ -58,17 +58,29 @@ pub struct OpenbookV2PlaceOrder<'info> {
...
+
+    /// CHECK: The oracle can be one of several different account types
+    #[account(address = payer_bank.load()?.oracle)]
+    pub payer_oracle: UncheckedAccount<'info>,
 
     /// CHECK: The oracle can be one of several different account types
-    #[account(address = bank.load()?.oracle)]
-    pub oracle: UncheckedAccount<'info>,
+    #[account(address = receiver_bank.load()?.oracle)]
+    pub receiver_oracle: UncheckedAccount<'info>,
```
ckamm, line 83: Sorry for not being clearer, imo:
- explicitly pass the payer_bank and receiver_bank, because they need to be writable - and normally health accounts in remaining_accounts are not writable
- the receiver_vault is not needed and does not need to be passed
- it's more future proof to read the oracle prices from the health accounts with the account_retriever (like you are doing below), passing them explicitly here will need to be removed in the future anyway, so better drop receiver_oracle and payer_oracle from the explicit account list

programs/mango-v4/src/state/mango_account.rs
```diff
@@ -702,7 +702,10 @@ impl MangoAccountDynamicHeader {
     }
 
     pub fn expected_health_accounts(&self) -> usize {
-        self.token_count() * 2 + self.serum3_count() + self.perp_count() * 2
+        self.token_count() * 2
+            + self.serum3_count()
+            + self.perp_count() * 2
+            + self.openbook_v2_count() * 2
```
ckamm, line 708: why `* 2`? Isn't there only a single openbook_v2 account per active market?

programs/mango-v4/src/instructions/openbook_v2_cancel_all_orders.rs
riordanp: I son't think we can use the group as signer, since the OpenOrdersAccount owner is used in the seeds for the OpenOrdersIndexer. we can't share one indexer for the entire group

programs/mango-v4/src/instructions/openbook_v2_liq_force_cancel_orders.rs
```diff
@@ -0,0 +1,223 @@
...
+    );
+
+    // todo-pan: maybe allow passing side for cu opt
+    openbook_v2::cpi::cancel_all_orders(cpi_ctx, None, limit)
+}
+
+fn cpi_settle_funds(ctx: &OpenbookV2LiqForceCancelOrders, seeds: &[&[&[u8]]]) -> Result<()> {
+    let group = ctx.group.load()?;
+    let cpi_accounts = SettleFunds {
+        penalty_payer: ctx.payer.to_account_info(),
```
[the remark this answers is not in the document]
riordanp: the penalty is currently hard coded to zero https://github.com/openbook-dex/openbook-v2/blob/master/programs/openbook-v2/src/state/market.rs#L16

yeah that might be a problem if it were used, not sure there's a way around it though

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -0,0 +1,538 @@
...
+            openbook_market_external.quote_lot_size,
+        )
+    };
+
+    // Provide a readable error message in case the vault doesn't have enough tokens
+    let base_lot_size;
+    let quote_lot_size;
+    {
+        base_lot_size = openbook_market_external.base_lot_size;
+        quote_lot_size = openbook_market_external.quote_lot_size;
```
ckamm, line 125: minor: move the base_lot_size assignments into the `let` declarations. The whole `base/quote_lot_size` constants could be set after openbook_market_external was loaded and then possibly used instead of the full `openbook_market_external.base_lot_size` (right now some places use one and some the other)

programs/mango-v4/src/serum3_cpi.rs
```diff
@@ -155,6 +156,22 @@ impl OpenOrdersSlim {
             referrer_rebates_accrued: oo.referrer_rebates_accrued,
         }
     }
+    pub fn from_oo_v2(oo: &OpenOrdersV2, base_lot_size: i64, quote_lot_size: i64) -> Self {
+        let base_locked_native: u64 = (oo.position.bids_base_lots * base_lot_size)
```
ckamm, line 160: unfortunate that these are i64

programs/mango-v4/src/serum3_cpi.rs
```diff
@@ -155,6 +156,22 @@ impl OpenOrdersSlim {
...
+            .unwrap();
+        let quote_locked_native: u64 = (oo.position.asks_base_lots * quote_lot_size)
+            .try_into()
+            .unwrap();
+
+        Self {
+            native_coin_free: oo.position.base_free_native,
+            native_coin_total: base_locked_native + oo.position.base_free_native,
+            native_pc_free: oo.position.quote_free_native,
+            native_pc_total: quote_locked_native + oo.position.base_free_native,
```
ckamm, line 171: quote_free_native

programs/mango-v4/src/accounts_ix/openbook_v2_place_order.rs
```diff
@@ -50,27 +52,24 @@ pub struct OpenbookV2PlaceOrder<'info> {
...
-    /// The bank vault that pays for the order, if necessary
+    /// The bank vault that pays for the order
     #[account(mut)]
     pub payer_vault: Box<Account<'info, TokenAccount>>,
 
-    /// CHECK: The oracle can be one of several different account types
-    #[account(address = payer_bank.load()?.oracle)]
-    pub payer_oracle: UncheckedAccount<'info>,
+    /// The bank that receives the funds upon settlement
+    // token_index and payer_bank.vault == payer_vault is validated inline at #3
```
ckamm, line 70: outdated comment

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -0,0 +1,538 @@
...
+    });
+
+    ctx.accounts.payer_vault.reload()?;
+    let after_vault = ctx.accounts.payer_vault.amount;
+
+    // Placing an order cannot increase vault balance
+    require_gte!(before_vault, after_vault);
+
+    let mut payer_bank = ctx.accounts.payer_bank.load_mut()?;
+    let mut receiver_bank = ctx.accounts.receiver_bank.load_mut()?;
```
ckamm, line 235: need to verify that these accounts are the right ones, check the token_index for both (anchor this as `#4` and mention it in the comment on payer_bank and receiver_bank in the accounts struct)

Technically these indexes are assert'ed in update_bank_potential_tokens() but I'd prefer if the check was explicit here.

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -0,0 +1,538 @@
...
+        base_lot_size = openbook_market_external.base_lot_size;
+        quote_lot_size = openbook_market_external.quote_lot_size;
+        let max_base_lots: u64 = order.max_base_lots.try_into().unwrap();
+        let max_quote_lots: u64 = order.max_quote_lots_including_fees.try_into().unwrap();
+
+        let needed_amount = match order.side {
+            OpenbookV2Side::Ask => {
+                (max_base_lots * base_lot_size as u64).saturating_sub(before_oo.native_base_free())
+            }
+            OpenbookV2Side::Bid => max_quote_lots.saturating_sub(before_oo.native_quote_free()),
```
ckamm, line 133: this likely needs the quote_lot_size applied

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
riordanp: does the check in `#3` not cover this by checking against the market token indexes?

programs/mango-v4/src/serum3_cpi.rs
riordanp: I changed it to use u64, think it's a bit nicer

programs/mango-v4/src/instructions/openbook_v2_cancel_all_orders.rs
ckamm: Ah, I see!

programs/mango-v4/src/serum3_cpi.rs
```diff
@@ -155,6 +156,20 @@ impl OpenOrdersSlim {
             referrer_rebates_accrued: oo.referrer_rebates_accrued,
         }
     }
+    pub fn from_oo_v2(oo: &OpenOrdersV2, base_lot_size: u64, quote_lot_size: u64) -> Self {
+        let bids_base_lots: u64 = oo.position.bids_base_lots.try_into().unwrap();
+        let asks_base_lots: u64 = oo.position.asks_base_lots.try_into().unwrap();
+        let base_locked_native = bids_base_lots * base_lot_size;
+        let quote_locked_native = asks_base_lots * quote_lot_size;
```
ckamm, line 163: This sounds like it's the wrong side around: For bids we need to lock up quote and for asks we need to lock up base.

Also, bids_base_lots and asks_base_lots are both in _base_ units, meaning that to find the amount of quote locked up, we need to multiply with the price. Check out OpenOrdersAccount::cancel_order and use the same logic as there.

programs/mango-v4/src/serum3_cpi.rs
```diff
@@ -155,6 +156,20 @@ impl OpenOrdersSlim {
...
+
+        Self {
+            native_coin_free: oo.position.base_free_native,
+            native_coin_total: base_locked_native + oo.position.base_free_native,
+            native_pc_free: oo.position.quote_free_native,
+            native_pc_total: quote_locked_native + oo.position.quote_free_native,
+            referrer_rebates_accrued: oo.position.referrer_rebates_available,
```
ckamm, line 170: What do we do about `locked_maker_fees`? It sounds like they are locked when a bid is placed; consumed on fill and returned on cancel?

If so, it sounds like they need to be accounted for (I don't think they are for serum3, but it's more important in openbookv2 since market creators can decide fees, as far as I'm aware). When we compute health for the "all asks execute" and "all bids execute" cases, that's the place where fees either got consumed or reimbursed.

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -67,7 +67,7 @@ pub fn openbook_v2_place_order(
     let mut account = ctx.accounts.account.load_full_mut()?;
     let retriever = new_fixed_order_account_retriever(ctx.remaining_accounts, &account.borrow())?;
 
-    let (_, _, payer_active_index) = account.ensure_token_position(receiver_token_index)?;
+    let (_, _, payer_active_index) = account.ensure_token_position(payer_token_index)?;
```
ckamm, line 70: good catch!

programs/mango-v4/src/state/bank.rs
```diff
@@ -204,6 +204,10 @@ pub struct Bank {
     /// zero means none, in token native
     pub deposit_limit: u64,
 
+    /// Largest amount of tokens that might be added the the bank based on
+    /// oenbook open order execution.
+    pub potential_openbook_tokens: u64,
```
ckamm, line 209: it's too dangerous to add new fields that are not at the end, even after merges. Please move this field to the end, just before `reserved`

programs/mango-v4/src/health/cache.rs
```diff
@@ -314,7 +314,7 @@ impl SpotInfo {
             quote_info_index,
             serum3_market_index: None,
             openbook_v2_market_index: Some(open_orders.market_index),
-            has_zero_funds: open_orders_account.position.is_empty(),
+            has_zero_funds: open_orders_account.position.is_empty(1),
```
ckamm, line 317: prefer passing open_orders_account.version, to make this less mysterious

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -165,8 +165,11 @@ pub fn openbook_v2_place_order(
         openbook.highest_placed_ask = 0.0;
     }
     // in the normal quote per base units
-    let limit_price =
-        order.max_quote_lots_including_fees as f64 * quote_lot_size as f64 / base_lot_size as f64;
+    let bids = ctx.accounts.bids.load_mut()?;
+    let asks = ctx.accounts.asks.load_mut()?;
+    let order_book = openbook_v2::state::Orderbook { bids, asks };
+    let price_lots = order.price(now_ts, None, &order_book)?.0;
```
ckamm, line 171: this depends on the state of the orderbook and we've just modified it by placing an order; I think it'd be safer to figure out price_lots before any orderbook modification happens

also move down the comment "// in the normal quote per base units", it should cover `let limit_price = ...`

programs/mango-v4/src/serum3_cpi.rs
ckamm: (left this open because I want to double-check against the obv2 code)

programs/mango-v4/src/lib.rs
riordanp: it cannot! re-created those enums that appear in the interface, in our codebase to fix this

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
```diff
@@ -136,6 +136,12 @@ pub fn openbook_v2_place_order(
         });
     }
 
+    // Get price lots before the book gets modified
+    let bids = ctx.accounts.bids.load_mut()?;
+    let asks = ctx.accounts.asks.load_mut()?;
+    let order_book = openbook_v2::state::Orderbook { bids, asks };
+    let price_lots = order.price(now_ts, None, &order_book)?.0;
```
ckamm, line 143: we do explicitly not allow oracle pegged orders via mango, right? (this None will cause an error)

ts/client/src/accounts/group.ts
```diff
@@ -437,13 +437,19 @@ export class Group {
         }),
       );
     } else {
+      const pKs = Array.from(this.openbookV2MarketsMapByExternal.values()).map(
+        (m) => m.openbookMarketExternal,
```
ckamm, line 441: minor: why not use keys?

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
ckamm: no, the checks at `#3` don't really check what they should:
```
    let (receiver_bank, receiver_bank_oracle) =
        retriever.bank_and_oracle(&group_key, receiver_active_index, receiver_token_index)?;
    require_eq!(receiver_bank.token_index, receiver_token_index);
```
This just checks that `bank_and_oracle()` did its job (which read the bank from the health account list) (and this require_eq should likely be removed). Replace that check with `ctx.accounts.receiver_bank.load()?.token_index == receiver_token_index`, i.e. that the explicitly-passed bank is for the right token.

programs/mango-v4/src/accounts_ix/openbook_v2_settle_funds.rs
```diff
@@ -73,4 +73,5 @@ pub struct OpenbookV2SettleFunds<'info> {
     pub base_oracle: UncheckedAccount<'info>,
```
ckamm, line 73: the comment on these oracles I don't get "in the parent"? Probably a c&p error. Also, I didn't find validation, please add!

programs/mango-v4/src/instructions/openbook_v2_place_order.rs
riordanp: yes, we build the openbook v2 order in lib.rs:1769 from our own types so it's not possible to place oracle pegged orders

ts/client/src/accounts/group.ts
riordanp: could do, but they are strings so would have to map anyway to get pubkeys for the rpc call method

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.69"
+channel = "1.70"
```
ckamm, line 2: I wonder if this change is really needed? If it is, the ci jobs and dockefile also need to update to this rust version.

ts/client/scripts/idl-compare.ts
```diff
@@ -359,7 +359,7 @@ function typeSize(idl: Idl, ty: IdlType): number {
       return 32;
     default:
       if ('vec' in ty) {
-        return 4;
+        return 1;
```
ckamm, line 362: I don't think this is right, is it necessary?

rust-toolchain.toml
ckamm: I did get a stack overflow in cases::test_openbook_v2::test_openbook_compute with 1.69, so seems needed.

ckamm (commented):
The liquidator needs to be updated to also send OpenbookV2LiqForceCancelOrders

programs/mango-v4/src/instructions/openbook_v2_cancel_order.rs
```diff
@@ -0,0 +1,99 @@
...
+
+pub fn openbook_v2_cancel_order(
+    ctx: Context<OpenbookV2CancelOrder>,
+    side: OpenbookV2Side,
+    order_id: u128,
+) -> Result<()> {
+    // Check instruction gate
+    let group = ctx.accounts.group.load()?;
+    require!(
+        group.is_ix_enabled(IxGate::OpenbookV2CancelAllOrders),
```
ckamm, line 23: bad ix gate check

ckamm (approved):
Pending issues:
- fix bug in openbook v2 program, update fixture, uncomment test code
- test liquidator, maybe write obv2_liq_force_cancel program test

PR #838, opened 2024-01-04
Use fallback oracles in Rust client

Lou-Kamades:


lib/client/src/context.rs
```diff
@@ -554,6 +628,92 @@ impl MangoGroupContext {
...
+                }
+                stale_oracles_with_fallbacks.into_iter().collect()
+            }
+        };
+
+        Ok(fallbacks_by_oracle)
+    }
+
+    /// Returns the a map of oracle pubkey -> (fallback_oracle_key, fallback_oracle_quote_key).
+    pub fn derive_fallback_oracle_keys_sync(
```
Lou-Kamades, line 686: I hate to have an async and sync version, but the async version was causing thread safety issues in the settler. Specifically this thread: https://github.com/blockworks-foundation/mango-v4/blob/53bc52efe7c0b3eee01d4737be1f043c43cf9980/bin/settler/src/main.rs#L316 would throw `rust could not prove `[async block@bin/settler/src/main.rs:320:9: 335:10]: std::marker::Send`

Making `token_withdraw_instructions` async here: https://github.com/blockworks-foundation/mango-v4/blob/53bc52efe7c0b3eee01d4737be1f043c43cf9980/bin/settler/src/tcs_start.rs#L118 is what causes the issue, but I am still unable to understand why that's the case? Ultimatley I think something about the async version of derive_fallback_oracle_keys should be to blame but I can't pinpoint it.

Here's a similar Rust issue for more context: https://github.com/rust-lang/rust/issues/102211

ckamm: Yeah, those are a nightmare to debug. Basically what I did was to go through the callchain of the presumed culprit and checking if any of the context of the calls might not be Sent. After a bunch of staring and a few false leads, I noticed that the `.unique()` iterator function might very well return a non-send wrapper and made that into a vec to make it Send.

ckamm: A helpful thing I had noticed early was that derive_fallback_oracle_keys() wasn't the problem, it was token_withdraw_instructions() being async that caused the issue.

lib/client/src/client.rs
```diff
@@ -103,6 +107,7 @@ impl Client {
         commitment: CommitmentConfig,
         fee_payer: Arc<Keypair>,
         timeout: Option<Duration>,
+        fallback_oracle_config: Option<FallbackOracleConfig>,
```
ckamm, line 110: please don't change ::new() (unnecessary api change for users; and ::new() is deprecated over using a builder anyway)

bin/keeper/src/main.rs
```diff
@@ -103,6 +105,7 @@ async fn main() -> Result<(), anyhow::Error> {
                 commitment,
                 owner.clone(),
                 Some(Duration::from_secs(cli.timeout)),
+                Some(FallbackOracleConfig::Never),
```
ckamm, line 108: wouldn't it make sense for the keeper to support fallback oracles? a particular reason for Never here? (same question for the other bins and services)

bin/liquidator/src/liquidate.rs
```diff
@@ -541,9 +545,14 @@ pub async fn maybe_liquidate_account(
...
 
     let account = account_fetcher.fetch_mango_account(pubkey)?;
-    let health_cache = health_cache::new(&mango_client.context, account_fetcher, &account)
-        .await
-        .context("creating health cache 1")?;
+    let health_cache = health_cache::new(
+        &mango_client.context,
+        &mango_client.client.fallback_oracle_config,
+        account_fetcher,
+        &account,
```
ckamm, line 552: hmm, wonder if there could be a mango_client helper function to get the health cache for an account -- after all the context, fallback_oracle_config and account_fetcher are all available to the mango_client

lib/client/src/account_fetcher.rs
```diff
@@ -29,6 +29,8 @@ pub trait AccountFetcher: Sync + Send {
         program: &Pubkey,
         discriminator: [u8; 8],
     ) -> anyhow::Result<Vec<(Pubkey, AccountSharedData)>>;
+
+    fn rpc(&self) -> &RpcClientAsync;
```
ckamm, line 33: My intuiton is that this abstraction shouldn't require being tied to the particular RpcClientAsync. Looks like the only reason for it is that fetch_multiple isn't a primitive on the AccountFetcher trait - add that instead.

lib/client/src/chain_data_fetcher.rs
```diff
@@ -64,12 +67,34 @@ impl AccountFetcher {
 
     pub fn fetch_bank_and_price(&self, bank: &Pubkey) -> anyhow::Result<(Bank, I80F48)> {
         let bank: Bank = self.fetch(bank)?;
-        let oracle = self.fetch_raw(&bank.oracle)?;
-        let oracle_acc = &KeyedAccountSharedData::new(bank.oracle, oracle.into());
-        let price = bank.oracle_price(&OracleAccountInfos::from_reader(oracle_acc), None)?;
+        let oracle_data = self.fetch_raw(&bank.oracle)?;
+        let oracle = &KeyedAccountSharedData::new(bank.oracle, oracle_data.into());
```
ckamm, line 71: looks like `oracle` could also be fetched with the new helper?

lib/client/src/client.rs
```diff
@@ -428,13 +450,21 @@ impl MangoClient {
...
+        let fallback_contexts = self
+            .context
+            .derive_fallback_oracle_keys(
+                &self.client.fallback_oracle_config,
+                &self.client.rpc_async(),
+            )
+            .await?;
 
         let (health_check_metas, health_cu) =
             self.context.derive_health_check_remaining_account_metas(
```
ckamm, line 462: this sounds like there should be a new helper function for derive_health_check_remaining_account_metas in the MangoClient (similar how there's one for the health_check remainig account metas), that can hide fetching the fallback_contexts. Otherwise that needs to be repeated everywhere.

bin/keeper/src/main.rs
Lou-Kamades: My initial thought was to start with Never and we could upgrade as necessary -- I'll amend everything to the default (Dynamic)

lib/client/src/account_fetcher.rs
```diff
@@ -268,7 +282,14 @@ impl<T: AccountFetcher + 'static> AccountFetcher for CachedAccountFetcher<T> {
         }
     }
 
-    fn rpc(&self) -> &RpcClientAsync {
-        self.fetcher.rpc()
+    async fn fetch_multiple_accounts(
+        &self,
+        keys: &[Pubkey],
+    ) -> anyhow::Result<Vec<(Pubkey, AccountSharedData)>> {
+        self.fetcher.fetch_multiple_accounts(keys).await
```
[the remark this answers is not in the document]
Lou-Kamades: I implemented `fetch_multiple_accounts` naively, so it will not fetch accounts it already has. Instead, the keeper has been configured to never use fallback oracles (otherwise there would be much trouble with the staleness checks)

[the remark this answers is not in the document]
ckamm: I don't think it's good to use the first requested pubkey as a key for the job. There could be difficult to find bugs when there are multiple fetch_multiple_accounts running with different lists but the same first key.

I think we either just race and allow multiple fetch jobs to run at the same time or we do something like sort+hash the account keys that are being fetched.

Since multiple fetch_multiple queries with slightly different lists would run in parallel anyway, I think it's ok to give up and just
- check cache, return if all there
- start the job for the missing keys
- put them in the cache on success
- return

lib/client/src/account_fetcher.rs
```diff
@@ -303,7 +306,13 @@ impl<T: AccountFetcher + 'static> AccountFetcher for CachedAccountFetcher<T> {
...
 
             let self_copy = self.clone();
-            let job_key = missing_keys[0];
+            missing_keys.sort();
+            let job_key = missing_keys
+                .clone()
+                .into_iter()
+                .fold(Hash::from([0u8; HASH_BYTES]), |acc, key| {
+                    hashv(&[&acc.to_bytes(), &key.to_bytes()])
+                });
```
ckamm, line 315: very minor, since this works
- just .iter() instead of .clone().into_iter() should work?
- Hash::default() should work over the explicit creation
- I'd probably write
```
let hasher = Hasher::default();
for key in missing_keys {
    hasher.hash(&key);
}
let job_key = hasher.result();
```
right now you're computing a new hash for each.

ckamm (approved):

PR #843, opened 2024-01-09
use fast clone

abrzezinski94:
This will drop render of some elements on ui from 1.5 sec to few ms  

- To copy something more complicated use fast-copy lib its faster then lodash implementation - e.g group copy
- for health obj its good to use own implementation because we don't need to loop over BN or I80F48 with functions from libs.

Own implementation still do deep copy without reference of given obj.

### DIFF
--- a/ts/client/src/utils.ts
+++ b/ts/client/src/utils.ts
@@ -218,3 +218,36 @@ export declare abstract class As<Tag extends keyof never> {
   private static readonly $as$: unique symbol;
   private [As.$as$]: Record<Tag, true>;
 }
+
+export function deepClone<T>(obj: T, hash = new WeakMap()): T {
+  // Handle non-object types and functions
+  if (typeof obj !== 'object' || obj === null) return obj;
+
+  // Handle circular references
+  if (hash.has(obj)) return hash.get(obj) as T;
+
+  let result: any;
+
+  if (obj instanceof Map) {
+    result = new Map();
+    hash.set(obj, result);
+    obj.forEach((value, key) => {
+      result.set(deepClone(key, hash), deepClone(value, hash));
+    });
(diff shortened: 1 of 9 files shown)


ts/client/src/utils.ts
```diff
@@ -218,3 +218,36 @@ export declare abstract class As<Tag extends keyof never> {
   private static readonly $as$: unique symbol;
   private [As.$as$]: Record<Tag, true>;
 }
+
+export function deepClone<T>(obj: T, hash = new WeakMap()): T {
```
microwavedcola1, line 222: do you have a reference for this copy paste?

microwavedcola1: is it from stackoverflow?

abrzezinski94: Nope combined few methods from multiple sources when looking around for solution.

microwavedcola1 (approved):

ts/client/src/utils.ts
mschneider: why is it faster than lodash?

abrzezinski94: Health Objects are mostly I80F48 or BN, lodash is well optimized (fast-clone lib is faster added it also in this pr for group object cloning) and have custom implementation to cover corner cases, it will run more operations on objects like BN to clone them because it treat it as more complex then it really is. So when you just want to get rid of reference to do some math on those its enough to run native js functions to copy them i also made performance checks and its indeed faster. Also lodash don't preserve prototype chain like this method has but in case of BN it won't affect math operations it would be problem when we would have more nested classes inside this object with functions that change values but for that case fast-copy is there.

https://www.npmjs.com/package/fast-copy have table for performance of different libs also runned my own with react tools.

PR #846, opened 2024-01-11
Improve tx confirmation

abrzezinski94:
currently there is problem that sometimes tx is confirmed on solana exploler but our confirmation do not confirm this will fix that combination of 3 methods will give good results

### DIFF
--- a/ts/client/src/utils/rpc.ts
+++ b/ts/client/src/utils/rpc.ts
@@ -19,6 +19,7 @@ import {
 } from '@solana/web3.js';
 import { COMPUTE_BUDGET_PROGRAM_ID } from '../constants';
 import { TxCallbackOptions } from '../client';
+import { awaitTransactionSignatureConfirmation } from '@blockworks-foundation/mangolana/lib/transactions';
 
 export interface MangoSignatureStatus {
   confirmations?: number | null;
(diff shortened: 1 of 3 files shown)


microwavedcola1 (approved):

yarn.lock
```diff
@@ -302,24 +316,24 @@
...
+    agentkeepalive "^4.5.0"
     bigint-buffer "^1.1.5"
     bn.js "^5.2.1"
     borsh "^0.7.0"
     bs58 "^4.0.1"
     buffer "6.0.3"
     fast-stable-stringify "^1.0.0"
     jayson "^4.1.0"
-    node-fetch "^2.6.12"
+    node-fetch "^2.7.0"
```
mschneider, line 336: i'm not sure this is good, there was a reason why this is supposed to be pinned to 2.6.11 @ckamm

ts/client/src/utils/rpc.ts
```diff
@@ -171,20 +172,47 @@ const confirmTransaction = async (
...
+    status = await Promise.race([
+      connection.confirmTransaction(
+        {
+          signature: signature,
+          blockhash: latestBlockhash.blockhash,
+          lastValidBlockHeight: latestBlockhash.lastValidBlockHeight,
+        },
+        txConfirmationCommitment,
+      ),
+      awaitTransactionSignatureConfirmation({
```
mschneider, line 184: this should also check the backup connections

mschneider: https://github.com/blockworks-foundation/mango-v4/blob/53bc52efe7c0b3eee01d4737be1f043c43cf9980/ts/client/src/utils/rpc.ts#L139

yarn.lock
abrzezinski94: i think its because mangolana uses newest version of web3.js let me know what was the issues before with upgrading it i can look into that.

abrzezinski94: didn't see problems so ill merge it now want to test it live

PR #849, opened 2024-01-15
Add a token-token platform liquidation fee

ckamm:


programs/mango-v4/src/instructions/token_register_trustless.rs
```diff
@@ -105,7 +105,9 @@ pub fn token_register_trustless(
         fallback_oracle: ctx.accounts.fallback_oracle.key(),
         deposit_limit: 0,
         zero_util_rate: I80F48::ZERO,
-        reserved: [0; 1952],
+        platform_liquidation_fee: I80F48::ZERO, // ### TODO
+        collected_liquidation_fees: I80F48::ZERO,
```
mschneider, line 109: worth updating default liquidation fee to 0.1 as well at the same time

PR #852, opened 2024-01-17
use deposit limits

abrzezinski94:


ts/client/src/accounts/group.ts
```diff
@@ -604,6 +604,27 @@ export class Group {
...
+   * @param mintPk
+   * @returns remaining deposit limit for mint, returns null if there is no limit for bank
+   */
+  public getRemainingDepositLimitByMint(mint: PublicKey): BN | null {
+    const bank = this.getFirstBankByMint(mint);
+    const vaultBalance = this.getTokenVaultBalanceByMint(mint);
+    const isNoLimit = bank.depositLimit.isZero();
+
+    const remainingDepositLimit = !isNoLimit
+      ? bank.depositLimit.sub(vaultBalance)
```
microwavedcola1, line 618: use bank.nativeDeposits() and not vault balance

microwavedcola1: and then you can move this method to the bank, something like getRemainingDepositLimit

ts/client/src/accounts/mangoAccount.ts
```diff
@@ -607,6 +613,33 @@ export class MangoAccount {
...
+  public calculateEquivalentSourceAmount(
+    sourceBank: Bank,
+    targetBank: Bank,
+    targetRemainingDepositLimit: BN,
+  ): BN {
+    const decimalDifference = sourceBank.mintDecimals - targetBank.mintDecimals;
+    let adjustedExchangeRate: number;
+
+    // Adjust the exchange rate based on the decimal difference
+    if (decimalDifference > 0) {
```
microwavedcola1, line 625: IMO all this is not needed, you can simply replace this whole method by `targetRemainingDepositLimit .mul(targetBank.price.div(sourceBank.price))`, this should take care of native native conversion, also you shouldnt need to do the if else

PR #864, opened 2024-01-30
fix decoding of broken oracle

abrzezinski94:
currently if oracle never reported the price or something with the switchboard oracle is broken, client is starting to break bots and ui. 

Imo it should be catched inside ui or bot that the price is 0 and handled there depending on case, now one broken token breaks use of other tokens as well.

### DIFF
--- a/ts/client/src/accounts/oracle.ts
+++ b/ts/client/src/accounts/oracle.ts
@@ -93,8 +93,10 @@ export function switchboardDecimalToBig(sbDecimal: {
 export function parseSwitchboardOracleV2(
   program: SwitchboardProgram,
   accountInfo: AccountInfo<Buffer>,
+  oracle: PublicKey,
 ): { price: number; lastUpdatedSlot: number; uiDeviation: number } {
   try {
+    //
     const price = program.decodeLatestAggregatorValue(accountInfo)!.toNumber();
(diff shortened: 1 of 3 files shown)


ts/client/src/accounts/oracle.ts
```diff
@@ -94,17 +94,20 @@ export function parseSwitchboardOracleV2(
...
+    const stdDeviation = switchboardDecimalToBig(
+      program.decodeAggregator(accountInfo).latestConfirmedRound.stdDeviation,
+    );
 
-  if (!price || !lastUpdatedSlot)
-    throw new Error('Unable to parse Switchboard Oracle V2');
-  return { price, lastUpdatedSlot, uiDeviation: stdDeviation.toNumber() };
+    return { price, lastUpdatedSlot, uiDeviation: stdDeviation.toNumber() };
+  } catch (e) {
+    console.log('Unable to parse Switchboard Oracle V2', e);
```
microwavedcola1, line 108: Can you put the comment that there were 0 switchboard rounds, hence this happened

microwavedcola1, line 108: can you log also the pubkey?

microwavedcola1 (approved):

PR #866, opened 2024-02-01
rust client: optionally derive prio fees from feed

ckamm:
This allows connecting to a lite-rpc feed to receive block priority updates and computing recently good priority fee values based on that.

### DIFF
--- a/lib/client/src/priority_fees.rs
+++ b/lib/client/src/priority_fees.rs
@@ -1,0 +1,240 @@
+use futures::{SinkExt, StreamExt};
+use jsonrpc_core::{MethodCall, Notification, Params, Version};
+use std::collections::HashMap;
+use std::sync::{Arc, RwLock};
+use std::time::{Duration, Instant};
+use tokio::sync::broadcast;
+use tokio::task::JoinHandle;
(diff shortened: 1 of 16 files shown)


lib/client/src/priority_fees.rs
```diff
@@ -0,0 +1,240 @@
...
+    let notification: Notification = match serde_json::from_str(&notification_str) {
+        Ok(v) => v,
+        Err(_) => return Ok(None), // not a notification at all
+    };
+    if notification.method != "blockPrioritizationFeesNotification" {
+        return Ok(None);
+    }
+    let map = match notification.params {
+        Params::Map(m) => m,
+        _ => anyhow::bail!("unexpected params, expected map"),
```
Lou-Kamades, line 154: I'm trying to understand when exactly this case would be triggered -- is it only if the lite rpc endpoint returns some respone that does not adhere to the JSON-RPC spec?

ckamm: I didn't double check the spec, but jsonrpc_core also allows array params or null params. But the notifications we expect should always have an object as params.

microwavedcola1 (approved):

PR #867, opened 2024-02-02
liquidation: Disallow liquidating assets with zero maint asset weight

ckamm:
This removes the oracle dependency for tokens with zero maint asset weight that are not borrowable.

### DIFF
--- a/bin/liquidator/src/liquidate.rs
+++ b/bin/liquidator/src/liquidate.rs
@@ -370,20 +370,16 @@ impl<'a> LiquidateHelper<'a> {
             .filter_map(|(ti, effective)| {
                 // check constraints for liquidatable assets, see also has_possible_spot_liquidations()
                 let tokens = ti.balance_spot.min(effective.spot_and_perp);
-                let is_valid_asset = tokens >= 1;
+                let is_valid_asset = tokens >= 1 && ti.maint_asset_weight.is_positive();
                 let quote_value = tokens * ti.prices.oracle;
-                // prefer to liquidate tokens with asset weight that have >$1 liquidatable
-                let is_preferred =
-                    ti.init_asset_weight > 0 && quote_value > I80F48::from(1_000_000);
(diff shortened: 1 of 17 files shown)


microwavedcola1 (approved):

bin/liquidator/src/liquidate.rs
```diff
@@ -370,20 +370,16 @@ impl<'a> LiquidateHelper<'a> {
             .filter_map(|(ti, effective)| {
                 // check constraints for liquidatable assets, see also has_possible_spot_liquidations()
                 let tokens = ti.balance_spot.min(effective.spot_and_perp);
-                let is_valid_asset = tokens >= 1;
+                let is_valid_asset = tokens >= 1 && ti.maint_asset_weight.is_positive();
                 let quote_value = tokens * ti.prices.oracle;
-                // prefer to liquidate tokens with asset weight that have >$1 liquidatable
-                let is_preferred =
```
Lou-Kamades, line 376: we don't care about preferred assets anymore?

Lou-Kamades (approved):

bin/liquidator/src/liquidate.rs
ckamm: Indeed, previously the liquidator preferred to liquidate against assets that had asset weight (and at least $1 in deposits). Now that tokens with asset weight are the only ones that can be liquidated, the logic for it is no longer needed.

programs/mango-v4/src/health/cache.rs
farnyser, line 1: Shouldn't the new condition be added to `has_possible_spot_liquidations` ?

ckamm: Could you explain? https://github.com/blockworks-foundation/mango-v4/pull/867/files#diff-110f42497491be964f83fc2e8ddceda72faac825ae293a9d26f5bdc8b6635312R937 changes `has_possible_spot_liquidations()` to add the ` && ti.maint_asset_weight.is_positive()` for the asset side.

PR #868, opened 2024-02-02
Add collateral fees

ckamm:
- New permissionless instruction to regularly charge collateral fees
- Bank and group configuration to set rate and interval
- Keeper addition to call the instruction


programs/mango-v4/src/instructions/token_charge_collateral_fees.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        let is_active = bank.withdraw_without_fee(token_position, fee, now_ts)?;
+        if !is_active {
+            account.deactivate_token_position_and_log(raw_token_index, ctx.accounts.account.key());
+        }
+
+        bank.collected_fees_native += fee;
+        bank.collected_collateral_fees += fee;
+
+        // TODO: emit a log
```
microwavedcola1, line 73: hmm what to do in case charging fees pushes maint health below 0? that can happen right?

programs/mango-v4/src/instructions/token_charge_collateral_fees.rs
```diff
@@ -0,0 +1,77 @@
...
+        return Ok(());
+    }
+    account.fixed.last_collateral_fee_charge = now_ts;
+
+    // Charge the user at most for 2x the interval. So if no one calls this for a long time
+    // there won't be a huge charge based only on the end state.
+    let charge_seconds = (now_ts - last_charge_ts).min(2 * group.collateral_fee_interval);
+
+    let inv_seconds_per_day = I80F48::from_num(1.157407407407e-5); // 1 / (24 * 60 * 60)
+    let collateral_fee_scaling = I80F48::from(charge_seconds) * inv_seconds_per_day;
```
microwavedcola1, line 39: q: why this multiplication with inv? what is the usefulness of the inv?

programs/mango-v4/src/instructions/token_charge_collateral_fees.rs
ckamm: yes, that could happen - liquidators would notice and liquidate

programs/mango-v4/src/instructions/token_charge_collateral_fees.rs
ckamm: I assumed we configure the fees per day (similar to funding). So if we run this instruction once every 12-24h (example) then we need to compute how much of the daily fees apply each time. So it's `secs_since_last_time / seconds_per_day` to get a fraction of a day

microwavedcola1 (approved):

PR #870, opened 2024-02-07
Mc/tcs premium auction

microwavedcola1:


ts/client/src/client.ts
```diff
@@ -4339,9 +4325,11 @@ export class MangoClient {
           1) *
         100;
     }
-    const maxPricePremiumRate =
+    let maxPricePremiumRate =
       maxPricePremiumPercent > 0 ? maxPricePremiumPercent / 100 : 0.03;
 
+    maxPricePremiumRate = maxPricePremiumRate * 2;
```
ckamm, line 4331: Maybe make a separate helper for the 60s duration, 2x premium rate thing; I don't think this should be in this function

ts/client/src/client.ts
```diff
@@ -3756,8 +3752,9 @@ export class MangoClient {
     );
     const lowerLimit = 0;
     const upperLimit = thresholdPriceNativeNative;
+    console.log(thresholdPriceNativeNative);
```
ckamm, line 3755: logging

ts/client/src/client.ts
```diff
@@ -4339,8 +4339,10 @@ export class MangoClient {
           1) *
         100;
     }
-    const maxPricePremiumRate =
+    let maxPricePremiumRate =
       maxPricePremiumPercent > 0 ? maxPricePremiumPercent / 100 : 0.03;
+    maxPricePremiumRate =
+      maxPricePremiumRate * premiumMultiplier + extraPricePremiumBps / 10000;
```
ckamm, line 4345: Imo this modification of user-specified premium is weird. Would prefer if this was just done externally to the price premium argument. This function should just build the instruction with the passed-in args.

microwavedcola1: They do have defaults which act as no-op

PR #875, opened 2024-02-08
rust client: Perp - Auto close account when needed

farnyser:
Auto close first unused perp account _if_ it's needed to place a new order because all slots are currently allocated

This is done on the client side because we need health data to be correct on the program side when processing the place order

*Example tx:*
[3j5cZZcmc7MEXwNAsdoWKZRgea8SAXKs5gE2jxuXi2Pzgmxang9ZrgKwTL5fXPHaJchJ1e7wDNobJHSPBdEqpTz3](https://explorer.solana.com/tx/3j5cZZcmc7MEXwNAsdoWKZRgea8SAXKs5gE2jxuXi2Pzgmxang9ZrgKwTL5fXPHaJchJ1e7wDNobJHSPBdEqpTz3)

### DIFF
--- a/bin/cli/src/main.rs
+++ b/bin/cli/src/main.rs
@@ -1,4 +1,5 @@
 use clap::{Args, Parser, Subcommand};
+use mango_v4::state::{PlaceOrderType, SelfTradeBehavior, Side};
 use mango_v4_client::{
     keypair_from_cli, pubkey_from_cli, Client, MangoClient, TransactionBuilderConfig,
 };
@@ -6,7 +7,6 @@ use solana_sdk::pubkey::Pubkey;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
-use mango_v4::state::{PlaceOrderType, SelfTradeBehavior, Side};
 
 mod save_snapshot;
 mod test_oracles;
@@ -158,7 +158,7 @@ enum Command {
         #[clap(short, long)]
         output: String,
     },
-    PerpPlaceOrder(PerpPlaceOrder)
+    PerpPlaceOrder(PerpPlaceOrder),
 }
 
 impl Rpc {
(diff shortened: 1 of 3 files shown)


bin/cli/src/main.rs
```diff
@@ -88,6 +90,34 @@ struct JupiterSwap {
...
+
+    /// also pays for everything
+    #[clap(short, long)]
+    owner: String,
+
+    #[clap(long)]
+    market_name: String,
+
+    #[clap(long)]
+    side: String,
```
ckamm, line 106: just fyi, no need to change: `clap(value_enum)` is pretty nice for cases like this, see liquidator's main

programs/mango-v4/src/state/mango_account.rs
```diff
@@ -1196,6 +1197,19 @@ impl<
         Ok(())
     }
 
+    pub fn find_first_unused_perp_position(&self) -> Option<&PerpPosition> {
```
ckamm, line 1200: I think "unused" might be misunderstood as "inactive". Maybe `_active_unused_` even if it's verbose?

lib/client/src/client.rs
```diff
@@ -1057,58 +1057,107 @@ impl MangoClient {
...
+        perk_market_index: PerpMarketIndex,
+    ) -> anyhow::Result<Option<(PreparedInstructions, MangoAccountValue)>> {
+        let context = &self.context;
+        let settle_token_index = context.perp(perk_market_index).settle_token_index;
+
+        let mut account = account.clone();
+        let enforce_position_result =
+            account.ensure_perp_position(perk_market_index, settle_token_index);
+
+        if enforce_position_result.is_err() {
```
ckamm, line 1137: would prefer if this checked for NoFreePerpPositionIndex, in case enforce_perp_position() changes

minor: I also typically prefer to return early over indenting the remaining function

lib/client/src/client.rs
```diff
@@ -1057,58 +1057,107 @@ impl MangoClient {
...
+                    let previous_market = context.perp(perp_position_to_close.market_index);
+                    account.deactivate_perp_position(
+                        perp_position_to_close.market_index,
+                        previous_market.settle_token_index,
+                    )?;
+                    account.ensure_perp_position(perk_market_index, settle_token_index)?;
+
+                    Ok(Some((close_ix, account)))
+                }
+                None => Err(anyhow::format_err!("No perp market slot available")),
```
ckamm, line 1154: `anyhow::bail!(...)` is convenient here

ckamm (approved):

PR #877, opened 2024-02-12
Rust client: auto close spot account if needed to place a new order 

farnyser:
Example tx:
https://explorer.solana.com/tx/FNjZpELFjhqCh9sxcfUNt2YtNTRjkvkWvt7hguAQ4RePKHWcsmrv5PUAvngCECz2BmbRWpod2DESvg6myoKedZK

When placing a new serum3 order, if needed:
1/ try to close unused token to free slots for new order base/quote tokens
2/ try to close unused spot market to free slot for new order
3/ if 1 failed, try again (closing one market reduce token usage count)
4/ try to create the new spot market slot

### DIFF
--- a/bin/cli/src/main.rs
+++ b/bin/cli/src/main.rs
@@ -1,5 +1,6 @@
 use clap::clap_derive::ArgEnum;
 use clap::{Args, Parser, Subcommand};
+use mango_v4::accounts_ix::{Serum3OrderType, Serum3SelfTradeBehavior, Serum3Side};
 use mango_v4::state::{PlaceOrderType, SelfTradeBehavior, Side};
 use mango_v4_client::{
     keypair_from_cli, pubkey_from_cli, Client, MangoClient, TransactionBuilderConfig,
@@ -126,6 +127,34 @@ struct PerpPlaceOrder {
     rpc: Rpc,
 }
 
+#[derive(Args, Debug, Clone)]
+struct Serum3PlaceOrder {
+    #[clap(long)]
+    account: String,
+
+    /// also pays for everything
+    #[clap(short, long)]
+    owner: String,
+
+    #[clap(long)]
+    market_name: String,
+
+    #[clap(long, value_enum)]
+    side: CliSide,
+
+    #[clap(short, long)]
+    price: f64,
+
+    #[clap(long)]
+    quantity: f64,
+
+    #[clap(long)]
+    expiry: u64,
+
+    #[clap(flatten)]
+    rpc: Rpc,
+}
+
 #[derive(Subcommand, Debug, Clone)]
 enum Command {
     CreateAccount(CreateAccount),
@@ -167,6 +196,7 @@ enum Command {
         output: String,
     },
     PerpPlaceOrder(PerpPlaceOrder),
+    Serum3PlaceOrder(Serum3PlaceOrder),
 }
 
 impl Rpc {
@@ -329,6 +359,47 @@ async fn main() -> Result<(), anyhow::Error> {
                     },
                     10,
                     SelfTradeBehavior::AbortTransaction,
+                )
+                .await?;
+            println!("{}", txsig);
+        }
+        Command::Serum3PlaceOrder(cmd) => {
+            let client = cmd.rpc.client(Some(&cmd.owner))?;
+            let account = pubkey_from_cli(&cmd.account);
+            let owner = Arc::new(keypair_from_cli(&cmd.owner));
+            let client = MangoClient::new_for_existing_account(client, account, owner).await?;
+            let market_index = client.context.serum3_market_index(&cmd.market_name);
+            let market = client.context.serum3(market_index);
+            let base_token = client.context.token(market.base_token_index);
+            let quote_token = client.context.token(market.quote_token_index);
+
+            fn native(x: f64, b: u32) -> u64 {
+                (x * (10_i64.pow(b)) as f64) as u64
+            }
+
+            // coin_lot_size = base lot size ?
+            // cf priceNumberToLots
+            let price_lots = native(cmd.price, quote_token.decimals as u32) * market.coin_lot_size
+                / native(market.pc_lot_size as f64, base_token.decimals as u32);
+
+            // cf baseSizeNumberToLots
+            let max_base_lots =
+                native(cmd.quantity, base_token.decimals as u32) / market.coin_lot_size;
+
+            let txsig = client
+                .serum3_place_order(
+                    &cmd.market_name,
+                    match cmd.side {
+                        CliSide::Bid => Serum3Side::Bid,
+                        CliSide::Ask => Serum3Side::Ask,
+                    },
+                    price_lots,
+                    max_base_lots as u64,
+                    ((price_lots * max_base_lots) as f64 * 1.01) as u64,
+                    Serum3SelfTradeBehavior::AbortTransaction,
+                    Serum3OrderType::Limit,
+                    SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs(),
+                    10,
                 )
                 .await?;
             println!("{}", txsig);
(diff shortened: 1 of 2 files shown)


lib/client/src/client.rs
```diff
@@ -788,6 +898,142 @@ impl MangoClient {
...
+
+        for p in account.all_serum3_orders() {
+            let open_orders_acc = self
+                .account_fetcher
+                .fetch_raw_account(&p.open_orders)
+                .await?;
+            let open_orders_bytes = open_orders_acc.data();
+            let open_orders_data: &serum_dex::state::OpenOrders = bytemuck::from_bytes(
+                &open_orders_bytes[5..5 + std::mem::size_of::<serum_dex::state::OpenOrders>()],
+            );
```
ckamm, line 915: now that this is done twice (also in serum3_cancel_all_orders), could you make a helper that goes from pubkey to open orders data?

lib/client/src/client.rs
```diff
@@ -788,6 +898,142 @@ impl MangoClient {
...
+                && open_orders_data.native_pc_total == 0;
+
+            if is_closable {
+                serum3_closable_order_market_index = Some(p.market_index);
+                break;
+            }
+        }
+
+        let first_closable_slot =
+            serum3_closable_order_market_index.expect("couldn't find any serum3 slot available");
```
ckamm, line 928: `a.expect()` will panic the program if a is None. Prefer to convert this into an error that we can return

lib/client/src/client.rs
```diff
@@ -788,6 +898,142 @@ impl MangoClient {
...
+    ) -> anyhow::Result<Option<PreparedInstructions>> {
+        let token_position_missing = account
+            .ensure_token_position(token_index)
+            .is_anchor_error_with_code(MangoError::NoFreeTokenPositionIndex.error_code());
+
+        if !token_position_missing {
+            return Ok(None);
+        }
+
+        let ixs = self.deactivate_first_active_unused_token(account).await?;
```
ckamm, line 982: the behavior here is a bit inconsistent:
- if the token position could be created directly on `account`, it's being done
- if it can't be created, it may close a different position but wil _not_ alter `account` to add the new position

lib/client/src/client.rs
```diff
@@ -788,6 +898,142 @@ impl MangoClient {
...
+            let bank = self.first_bank(x.token_index).await?;
+            let amount = x.native(&bank);
+            let name = bank.name().to_string();
+            all_pos.push((x.token_index, x.in_use_count, amount, name));
+        }
+
+        let closable_tokens = account
+            .all_token_positions()
+            .enumerate()
+            .filter(|p| p.1.is_active() && !p.1.is_in_use());
```
ckamm, line 1002: just fyi, another way to write this would be `|(_, p)| p.is_active() && !p.is_in_use()` which I weakly prefer

lib/client/src/client.rs
```diff
@@ -788,6 +898,142 @@ impl MangoClient {
...
+
+        let mut closable_token_position_raw_index_opt = None;
+        let mut closable_token_bank_opt = None;
+
+        for (closable_token_position_raw_index, closable_token_position) in closable_tokens {
+            let bank = self.first_bank(closable_token_position.token_index).await?;
+            let deposit = closable_token_position.native(&bank);
+
+            if deposit > I80F48::ONE {
```
ckamm, line 1011: it's not really `deposit`, more like the native balance on that token position. Can be negative if the user has borrows and then can't be closed!

lib/client/src/client.rs
```diff
@@ -788,6 +898,142 @@ impl MangoClient {
...
+        &self,
+        account: &mut MangoAccountValue,
+    ) -> anyhow::Result<PreparedInstructions> {
+        let mut all_pos = Vec::new();
+        for x in account.all_token_positions() {
+            let bank = self.first_bank(x.token_index).await?;
+            let amount = x.native(&bank);
+            let name = bank.name().to_string();
+            all_pos.push((x.token_index, x.in_use_count, amount, name));
+        }
```
ckamm, line 997: all_pos isn't used later in the function?

lib/client/src/client.rs
```diff
@@ -1963,7 +2209,7 @@ impl MangoClient {
 #[derive(Debug, thiserror::Error)]
 pub enum MangoClientError {
     #[error("Transaction simulation error. Error: {err:?}, Logs: {}",
-        .logs.iter().join("; ")
+    .logs.iter().join("; ")
```
ckamm, line 2212: this one doesn't look nice - did cargo fmt do it?

lib/client/src/client.rs
```diff
@@ -721,24 +768,87 @@ impl MangoClient {
...
+        }
+
+        let open_orders = open_orders_opt.expect("oo is created");
 
         let (health_check_metas, health_cu) = self
-            .derive_health_check_remaining_account_metas(account, vec![], vec![], vec![])
+            .derive_health_check_remaining_account_metas(
+                &account,
+                vec![],
+                vec![receiver_token.token_index],
```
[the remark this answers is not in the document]
farnyser: I had to add the receiver as you told me it was needed for place_order_v2 (unlinked to my others changes)

lib/client/src/client.rs
farnyser: That was for debugging, thx for pointing it out

lib/client/src/client.rs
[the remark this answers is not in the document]
ckamm: right, my mistake. I thought this was about affected tokens, but it's writable banks.

bin/cli/src/main.rs
```diff
@@ -429,7 +429,8 @@ async fn main() -> Result<(), anyhow::Error> {
             // coin_lot_size = base lot size ?
             // cf priceNumberToLots
             let price_lots = native(cmd.price, quote_token.decimals as u32) * market.coin_lot_size
-                / native(market.pc_lot_size as f64, base_token.decimals as u32);
+                / native(1.0, base_token.decimals as u32)
+                * market.pc_lot_size;
```
ckamm, line 433: need parens around the denominator

lib/client/src/client.rs
```diff
@@ -666,37 +666,40 @@ impl MangoClient {
...
+                        owner: self.owner(),
+                        sol_destination: self.owner(),
+                    },
+                    None,
+                ),
+                data: anchor_lang::InstructionData::data(
+                    &mango_v4::instruction::Serum3CloseOpenOrders {},
+                ),
+            },
+            self.context.compute_estimates.health_cu_per_serum,
```
ckamm, line 695: now that I think about it, health_cu_per_serum isn't the right choice. Use cu_per_mango_instruction - it's a large overestimate but 6k cu is too little

lib/client/src/client.rs
```diff
@@ -779,65 +782,10 @@ impl MangoClient {
...
-                self.context.compute_estimates.health_cu_per_serum,
-            );
-
-            let created_open_orders = self.serum3_create_open_orders_address(market_index);
-            open_orders_opt = Some(created_open_orders);
-            account.create_serum3_orders(market_index)?.open_orders = created_open_orders;
-        }
-
-        let open_orders = open_orders_opt.expect("oo is created");
+            .expect("oo is created");
```
ckamm, line 788: don't use `expect()` unless you want to crash on error. This would be better with `?` I think

lib/client/src/client.rs
```diff
@@ -769,7 +772,7 @@ impl MangoClient {
         limit: u16,
     ) -> anyhow::Result<PreparedInstructions> {
         let mut ixs = PreparedInstructions::new();
-        let mut account = account.clone();
+        let account = account.clone();
```
ckamm, line 775: this clone should not be needed

lib/client/src/client.rs
farnyser: yes it's cargo fmt

lib/client/src/client.rs
farnyser: that's how it was before, but I can change it if you think it's better

ckamm: would prefer if you did, but then it's optional

ckamm (approved):

PR #878, opened 2024-02-13
tests: Check mango account backwards compatibility

ckamm:


programs/mango-v4/src/state/mango_account.rs
```diff
@@ -2862,4 +2864,89 @@ mod tests {
...
+        for fixture in fixtures {
+            let filename = format!("resources/test/{}.bin", fixture);
+            let account_bytes = read_file(find_file(&filename).unwrap());
+
+            // Try a roundtrip via borsh
+            {
+                let mut account_bytes_slice: &[u8] = &account_bytes;
+                let data = MangoAccount::try_deserialize(&mut account_bytes_slice)?;
+                let mut borsh_bytes = Vec::new();
+                data.try_serialize(&mut borsh_bytes)?;
```
farnyser, line 2889: do we actually need to serialize back in the same format ?
if we don't, maybe do this instead:

```old_serialized = read_file()
old_deserialized = deserialize(old_serialized)
new_serialized = serialized(old_deserialized)
new_deserialized = deserialize(new_serialized)
assert_eq(old_deserialized, new_deserialized)```

PR #884, opened 2024-02-15
Add force_withdraw state and instruction

ckamm:


programs/mango-v4/src/instructions/token_force_withdraw.rs
```diff
@@ -0,0 +1,85 @@
...
+    // even if the oracle is bad.
+    let oracle_ref = &AccountInfoRef::borrow(ctx.accounts.oracle.as_ref())?;
+    let unsafe_oracle_state = oracle_state_unchecked(
+        &OracleAccountInfos::from_reader(oracle_ref),
+        bank.mint_decimals,
+    )?;
+
+    // Update the net deposits - adjust by price so different tokens are on the same basis (in USD terms)
+    let amount_usd = (amount_i80f48 * unsafe_oracle_state.price).to_num::<i64>();
+    account.fixed.net_deposits -= amount_usd;
```
microwavedcola1, line 67: :D I would have completely missed this, and once we miss such things, hard to reconcile

ts/client/src/client.ts
```diff
@@ -625,6 +618,93 @@ export class MangoClient {
...
+          ),
+        );
+      }
+    } else {
+      const account = await unpackAccount(ownerAtaTokenAccount, ai);
+      // if owner is not same as mango account's owner, then create another token account
+      if (!account.owner.equals(mangoAccount.owner)) {
+        const kp = Keypair.generate();
+        alternateOwnerTokenAccount = kp.publicKey;
+        await createAccount(
```
microwavedcola1, line 669: I chose to not extract ixs, and just do the tx here, to avoid a lot of copy paste, and use the library provided function

microwavedcola1 (approved):

PR #887, opened 2024-02-19
Audit v0.22 fixes

ckamm:


microwavedcola1 (approved):

programs/mango-v4/src/state/mango_account_components.rs
```diff
@@ -570,7 +570,7 @@ impl PerpPosition {
             .unwrap();
         let upnl_abs = upnl.abs().ceil().to_num::<i64>();
         self.recurring_settle_pnl_allowance =
-            self.recurring_settle_pnl_allowance.max(0).min(upnl_abs);
+            self.recurring_settle_pnl_allowance.min(upnl_abs).max(0);
```
farnyser, line 573: Is that only cosmetics ? 

By the way, I see a lot of .min().max() in the code, maybe we could use .clamp() instead for better readability ?

ckamm: yeah, only cosmetics. Clamp: agree!

PR #889, opened 2024-02-20
liquidator: avoid logging same oracle error (same token) in loop

farnyser:


bin/liquidator/src/main.rs
```diff
@@ -552,6 +562,26 @@ impl LiquidationState {
...
+                    .oracle_errors
+                    .had_too_many_errors(LiqErrorType::Liq, &ti, Instant::now())
+                    .is_some()
+                {
+                    println!(
+                        "{:?} recording oracle error for token {} {}",
+                        chrono::offset::Utc::now(),
+                        ti_name,
+                        ti
+                    );
```
ckamm, line 576: why does it log if had_too_many_errors() is true? shouldn't this output the error if there have not been too many oracle issues for that ti?

bin/liquidator/src/liquidate.rs
```diff
@@ -652,8 +654,8 @@ impl<'a> LiquidateHelper<'a> {
 
 #[allow(clippy::too_many_arguments)]
 pub async fn maybe_liquidate_account(
-    mango_client: &MangoClient,
-    account_fetcher: &chain_data::AccountFetcher,
```
ckamm, line 656: why's this needed?

bin/liquidator/src/unwrappable_oracle_error.rs
```diff
@@ -0,0 +1,45 @@
...
+            use std::str::FromStr;
+            let ti_res = u16::from_str(
+                error_str.split("token index ").collect::<Vec<&str>>()[1]
+                    .split(',')
+                    .collect::<Vec<&str>>()[0],
+            );
+            let ti_name = error_str.split(" name:").collect::<Vec<&str>>()[1]
+                .split(',')
+                .collect::<Vec<&str>>()[0]
+                .to_string();
```
ckamm, line 36: if we have to unpack the error_str, maybe nicer to use regex? see  programs/mango-v4/tests/cases/test_benchmark.rs for an example

bin/liquidator/src/unwrappable_oracle_error.rs
```diff
@@ -0,0 +1,126 @@
...
+
+            if !is_oracle_error {
+                return None;
+            }
+
+            let error_str = ae.to_string();
+            return parse_oracle_error_string(&error_str);
+        }
+
+        return None;
```
ckamm, line 32: very minor, but the value of the last statement in a function is returned automatically if there's no semicolon, so ideomatically you'd just write `None` here

ckamm (approved):

PR #890, opened 2024-02-20
service-mango-health: monitors/historizes accounts health

farnyser:


bin/service-mango-health/src/processors/persister.rs
```diff
@@ -0,0 +1,392 @@
...
+    async fn insert_history<'tx>(
+        client: &Transaction<'tx>,
+        updates: &HashMap<Pubkey, PersistedData>,
+    ) -> anyhow::Result<()> {
+        for (key, value) in updates {
+            let key = key.to_string();
+            let ts = value.computed_at.naive_utc();
+            let hr = value.health_ratio.map(|x| x.to_num::<f64>());
+            let query = postgres_query::query!("INSERT INTO mango_monitoring.health_history (Pubkey, Timestamp, HealthRatio) VALUES ($key, $ts, $hr)", key, ts, hr);
+            query.execute(client).await.expect("Insertion failed");
```
ckamm, line 156: for large number of inserts this is very inefficents, potential steps:
- use tokio_postgres directly, that way we can prepare() the insert statement ahead of time once
- I've seen people use `... VALUES (...row1...), (...row2...), ...` for like 10 rows to manually batch 10 row inserts at a time, but it's really awkward imo
- when looking around, I saw that tokio_postgres supports copy_in(). I think that's supposed to be the fastest way of getting lots of new rows into a table (`https://github.com/sfackler/rust-postgres/issues/681`)

At least the "prepare-statement-only-once" is an optimization I'd like to see.

bin/service-mango-health/src/processors/persister.rs
```diff
@@ -0,0 +1,392 @@
...
+    ) -> anyhow::Result<()> {
+        let to_update = updates
+            .into_iter()
+            .filter(|(key, _)| previous.contains_key(&key));
+
+        for (key, value) in to_update {
+            let key = key.to_string();
+            let ts = value.computed_at.naive_utc();
+            let hr = value.health_ratio.map(|x| x.to_num::<f64>());
+            let query = postgres_query::query!("UPDATE mango_monitoring.health_current SET Timestamp=$ts, HealthRatio=$hr WHERE Pubkey = $key", key, ts, hr);
```
ckamm, line 175: 1. Did you intentionally split INSERT and UPDATE? In my experience INSERT ON CONFLICT UPDATE works well.
2. Having to insert all this data a second time sounds expensive. Maybe health_current could be a materialized view of health_history? With the right index finding the most recent timestamp data for each pubkey should be fast?

bin/service-mango-health/src/processors/persister.rs
```diff
@@ -0,0 +1,392 @@
...
+        }
+
+        Ok(())
+    }
+
+    async fn delete_old_history<'tx>(
+        client: &Transaction<'tx>,
+        now: chrono::DateTime<Utc>,
+    ) -> anyhow::Result<()> {
+        let min_ts = (now - chrono::Duration::days(31)).naive_utc();
```
ckamm, line 206: I think this should be configurable

bin/service-mango-health/conf/example-config.toml
```diff
@@ -0,0 +1,26 @@
...
+mango_group = "78b8f4cGCwmZ9ysPFMWLaLTkkaYnUjwMJYStWe5RTSSX"
+
+# [postgres]
+# connection_string = "$PG_CONNECTION_STRING"
+# allow_invalid_certs = true
+# max_retry_count = 2
+
+# # [postgres.tls]
+# # ca_cert_path = "$PG_CA_CERT"
+# # client_key_path = "$PG_CLIENT_KEY"
```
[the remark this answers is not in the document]
farnyser: it's optional, can output accounts health to logs without postgres if needed

bin/service-mango-health/src/processors/persister.rs
```diff
@@ -0,0 +1,725 @@
...
+        let persistence_configuration = configuration.persistence_configuration.clone();
+        let time_to_live = Duration::seconds(persistence_configuration.history_time_to_live_secs);
+        let periodicity = Duration::seconds(persistence_configuration.persist_max_periodicity_secs);
+        let max_snapshot_count = persistence_configuration.snapshot_queue_length;
+
+        let mut unpersisted_snapshots = VecDeque::new();
+
+        let job = tokio::spawn(async move {
+            if !persistence_configuration.enabled {
+                return;
```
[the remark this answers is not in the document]
farnyser: This job exits immediately, but the logger should still keep going. 
Broadcast are allowed to any number of receivers (and it can change during run time)

ckamm (approved):

PR #893, opened 2024-02-24
Updated Docker Image File for Keeper/Liquidator

FinnCF:


ckamm (commented):
Please revert the `transaction_size()` change: it's goal is to just return the size. The check on number of accounts and bytes is done separately and I fixed the bug with how it was used in the keeper.

MANAALHARBI (approved):

MANAALHARBI (approved):

PR #894, opened 2024-02-26
add TokenBalanceLog in token_charge_collateral_fees

Lou-Kamades:


programs/mango-v4/src/logs.rs
```diff
@@ -803,6 +803,7 @@ pub struct TokenCollateralFeeLog {
     pub token_index: u16,
     pub asset_usage_fraction: i128,
     pub fee: i128,
+    pub price: i128,
```
ckamm, line 806: are you happy with changing the event definition like this? since v0.23 isn't released yet I'd say it's ok if it doesn't make things hard on the scraping side

ckamm (commented):
Looks good, need to measure CU per active token position again tho so the keeper allocates the right budget

ckamm:
Could you increase cu_per_charge_collateral_fees_token in context.rs from 12k to 15k? my measurements show:
- charge without any tokens: 13.3k CU (same as before)
- charge with one token: 25.9k CU (before: 25.4k CU)
So it doesn't increase CU much but my 12k guess was too optimistic.

programs/mango-v4/src/logs.rs
Lou-Kamades: yes please, this makes my life a lot easier

ckamm (approved):

PR #908, opened 2024-03-07
program: add a sequence check IX

farnyser:
This new IX `SequenceCheck` can be used to avoid having multiple concurrent TX in flight causing unexpected result (multiple borrow for example)


programs/mango-v4/src/instructions/sequence_check.rs
```diff
@@ -0,0 +1,17 @@
+use anchor_lang::prelude::*;
+
+use crate::accounts_ix::*;
+use crate::error::MangoError;
+use crate::state::*;
+
+pub fn sequence_check(ctx: Context<SequenceCheck>, expected_sequence_number: u64) -> Result<()> {
+    let mut account = ctx.accounts.account.load_full_mut()?;
+
+    require!(
```
ckamm, line 10: prefer require_eq

programs/mango-v4/tests/cases/test_basic.rs
```diff
@@ -947,3 +947,107 @@ async fn test_withdraw_skip_bank() -> Result<(), TransportError> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn test_sequence_check() -> Result<(), TransportError> {
+    // TODO FAS
```
ckamm, line 953: what's todo?

ckamm (approved):

ts/client/src/client.ts
```diff
@@ -1034,6 +1034,22 @@ export class MangoClient {
...
+    const ix = await this.program.methods
+      .sequenceCheck(mangoAccount.sequenceNumber)
+      .accounts({
+        group: group.publicKey,
+        account: mangoAccount.publicKey,
+        owner: (this.program.provider as AnchorProvider).wallet.publicKey,
+      })
+      .instruction();
+
+    return await this.sendAndConfirmTransactionForGroup(group, [ix]);
```
ckamm, line 1050: this would send a transaction with a single sequenceCheck ix. Instead, make this a function that returns the sequence check instruction, so it can be composed with other ixes.

ckamm (approved):

PR #909, opened 2024-03-10
program: sequence check as u8

farnyser:


mschneider (approved):
lgtm, small nit: could be inserted right after bump which is a 1 byte pad but not sure if that is guaranteed to be 0 @ckamm

ckamm (approved):

PR #914, opened 2024-03-12
Serge/liquidator split tcs and liquidation

farnyser:


bin/liquidator/src/main.rs
```diff
@@ -234,15 +245,15 @@ async fn main() -> anyhow::Result<()> {
         forbidden_tokens: liq_config.forbidden_tokens.clone(),
     };
 
-    let mut rebalance_interval = tokio::time::interval(Duration::from_secs(30));
-    let (rebalance_trigger_sender, rebalance_trigger_receiver) = async_channel::bounded::<()>(1);
+    let (rebalance_trigger_sender, rebalance_trigger_receiver) = tokio::sync::mpsc::channel(1);
```
ckamm, line 248: what's the reasoning for switching to tokio::sync::mpsc? in the past I had issues where tokio's channels where very bare-bones and didn't have functionality I wanted. Possibly that has changed since.

bin/liquidator/src/main.rs
```diff
@@ -370,126 +397,85 @@ async fn main() -> anyhow::Result<()> {
...
-                        let mut metric_tcs_start_end =
-                            metrics.register_latency("tcs_start_end".into());
-                        metric_tcs_start_end.push(current_time - tcs_start_time.unwrap());
-                        tcs_start_time = None;
-                    }
-                }
+    use cli_args::{BoolArg, Cli, CliDotenv};
+    use futures::StreamExt;
+    use liquidation_state::LiquidationState;
+    use tcs_state::TcsState;
```
ckamm, line 455: these `use` declarations are weird here

bin/liquidator/src/main.rs
```diff
@@ -370,126 +397,85 @@ async fn main() -> anyhow::Result<()> {
         }
     });
 
+    let mut optional_jobs = vec![];
+
     // Could be refactored to only start the below jobs when the first snapshot is done.
     // But need to take care to abort if the above job aborts beforehand.
+    if cli.rebalance == BoolArg::True {
```
ckamm, line 404: I like pulling this check out of the job

bin/liquidator/src/main.rs
```diff
@@ -234,15 +245,15 @@ async fn main() -> anyhow::Result<()> {
         forbidden_tokens: liq_config.forbidden_tokens.clone(),
     };
 
-    let mut rebalance_interval = tokio::time::interval(Duration::from_secs(30));
-    let (rebalance_trigger_sender, rebalance_trigger_receiver) = async_channel::bounded::<()>(1);
+    let (rebalance_trigger_sender, rebalance_trigger_receiver) = tokio::sync::mpsc::channel(1);
+    let (tx_trigger_sender, tx_trigger_receiver) = tokio::sync::mpsc::channel::<TxTrigger>(100);
```
ckamm, line 249: This is where I worry about the usual bounded channel semantic. We only care about the most recent, but this channel will block new generations when it's full. 

At some point we should write (a wrapper around an existing?) channel that drops the oldest entry when a new one is pushed to it. For now it can be done receiver-side, but will need to check that.

Just putting triggers in the channel makes it a lot less problematic, but still: could be full of "tcs" triggers and newer liquidation triggers can't be added.

Is it even helpful to have typed triggers? The only thing you want is to wake up a tx_sender thread, so it can pick up some available work if it was sleeping?

bin/liquidator/src/main.rs
```diff
@@ -370,126 +397,85 @@ async fn main() -> anyhow::Result<()> {
...
+    shared_state: Arc<RwLock<SharedState>>,
+    token_swap_info_updater: Arc<TokenSwapInfoUpdater>,
+) -> JoinHandle<()> {
+    tokio::spawn({
         let mut interval = mango_v4_client::delay_interval(Duration::from_secs(
             cli.token_swap_refresh_interval_secs,
         ));
         let mut startup_wait = mango_v4_client::delay_interval(Duration::from_secs(1));
-        let shared_state = shared_state.clone();
+        let shared_state = shared_state;
```
ckamm, line 478: looks odd

bin/liquidator/src/main.rs
farnyser: Didn't realized `async_channel` is multi producer "multi" consumer (single consume). 
Going back to that for rebalancing, and I'll probably update the PR to use it for TxTrigger also

bin/liquidator/src/tx_sender.rs
```diff
@@ -0,0 +1,130 @@
...
+                        // a task is supposed to be available to process
+                        // find it in global shared state, and mark it as processing
+                        // (it's also possible we don't find anything in case liq/tcs job enqueued the same task multiple time)
+                        let (l, t) = {
+                            let mut writer = shared_state.write().unwrap();
+
+                            // print out list of all task for debugging
+                            for x in &writer.liquidation_candidates_accounts {
+                                if !writer.processing_liquidation.contains(x) {
+                                    tracing::trace!(" - LIQ {:?}", x);
```
ckamm, line 45: minor, I often like to bring the info!/trace!/ etc macros into file scope; also you could write "{x:?}"

bin/liquidator/src/tx_sender.rs
```diff
@@ -0,0 +1,130 @@
...
+                                    tracing::trace!("  - TCS {:?}", x);
+                                }
+                            }
+
+                            // next task to execute
+                            let liq_candidate = writer
+                                .liquidation_candidates_accounts
+                                .iter()
+                                .find(|x| !writer.processing_liquidation.contains(x))
+                                .map(|x| *x);
```
ckamm, line 59: minor, you can use `.copied()` instead to turn an `Opt<&T> into an Opt<T>`

bin/liquidator/src/tx_sender.rs
```diff
@@ -0,0 +1,130 @@
...
+                                writer.processing_tcs.insert(t);
+                                (None, Some(t))
+                            } else {
+                                debug!("worker #{} got nothing", id);
+                                (None, None)
+                            }
+                        };
+
+                        // execute the task
+                        let done = if let Some(l) = l {
```
ckamm, line 86: minor, would avoid "l" as a variable name because of mistakes with "1"

lib/client/src/util.rs
```diff
@@ -20,19 +20,29 @@ impl<T, E: std::fmt::Debug> AnyhowWrap for Result<T, E> {
 /// Push to an async_channel::Sender and ignore if the channel is full
 pub trait AsyncChannelSendUnlessFull<T> {
     /// Send a message if the channel isn't full
-    fn send_unless_full(&self, msg: T) -> Result<(), async_channel::SendError<T>>;
+    fn send_unless_full(&self, msg: T) -> Result<(), anyhow::Error>;
```
ckamm, line 23: minor, prefer `anyhow::Result<()>` (which is the same type)

bin/liquidator/src/liquidation_state.rs
```diff
@@ -0,0 +1,238 @@
...
+
+                let found_candidates = liquidation
+                    .find_candidates(account_addresses.iter(), |p| {
+                        if shared_state
+                            .write()
+                            .unwrap()
+                            .liquidation_candidates_accounts
+                            .insert(p)
+                        {
+                            return tx_trigger_sender.send_unless_full(());
```
ckamm, line 214: this is misleading on an unbounded channel

bin/liquidator/src/tx_sender.rs
```diff
@@ -0,0 +1,130 @@
...
+                                debug!(
+                                    "worker #{} got a tcs candidate -> {:?} (out of {})",
+                                    id,
+                                    t,
+                                    writer.interesting_tcs.len()
+                                );
+                                writer.processing_tcs.insert(t);
+                                (None, Some(t))
+                            } else {
+                                debug!("worker #{} got nothing", id);
```
ckamm, line 80: this is indicative of a bug, isn't it? there's the assumption that each tx trigger channel message corresponds to one thing that needs to be done. If they don't match up it's possible that the channel is empty but there's still work to do.

bin/liquidator/src/main.rs
```diff
@@ -513,41 +491,56 @@ async fn main() -> anyhow::Result<()> {
...
+                }
+                if !shared_state.read().unwrap().one_snapshot_done {
+                    continue;
+                }
+                if let Err(err) = rebalancer.zero_all_non_quote().await {
+                    error!("failed to rebalance liqor: {:?}", err);
+
+                    // Workaround: We really need a sequence enforcer in the liquidator since we don't want to
+                    // accidentally send a similar tx again when we incorrectly believe an earlier one got forked
+                    // off. For now, hard sleep on error to avoid the most frequent error cases.
```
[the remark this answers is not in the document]
farnyser: Are you fine with me doing this in a separate pull-request ? I don't think it's more risky now even with the liq/tcs split

bin/liquidator/src/trigger_tcs.rs
```diff
@@ -1225,6 +1234,15 @@ impl Context {
             .instructions
             .append(&mut trigger_ixs.instructions);
 
+        let liqor = &self.mango_client.mango_account().await?;
+        tx_builder.instructions.append(
+            &mut self
+                .mango_client
+                .health_check_instruction(liqor, self.config.min_health_ratio, MaintRatio)
```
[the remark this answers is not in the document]
farnyser: I did it this way because you can actually check any account (permissionless)
But not sure why we would want to do that, so I'm okay to change that if you want ?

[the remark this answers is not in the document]
ckamm: both are fine, since all the uses right now pull the mango account before calling the function, I'd probably make a change - but you're right the current version is more flexible, so it's also ok to keep

bin/liquidator/src/liquidation_state.rs
```diff
@@ -0,0 +1,238 @@
...
+
+                let found_candidates = liquidation
+                    .find_candidates(account_addresses.iter(), |p| {
+                        if shared_state
+                            .write()
+                            .unwrap()
+                            .liquidation_candidates_accounts
+                            .insert(p)
+                        {
+                            return tx_trigger_sender.try_send(()).map_err(|e| e.into());
```
ckamm, line 214: probably could write it as `tx_trigger_sender.try_send(())?;`. Could also go with `send()` since the channel is unbounded...

bin/liquidator/src/tx_sender.rs
```diff
@@ -0,0 +1,178 @@
...
+            "worker #{} got a tcs candidate -> {:?} (out of {})",
+            id,
+            tcs_candidate,
+            writer.interesting_tcs.len()
+        );
+        writer.processing_tcs.insert(tcs_candidate);
+        return Ok(WorkerTask::Tcs(tcs_candidate));
+    }
+
+    Err(anyhow::format_err!("Worker #{} - No task found", id))
```
ckamm, line 150: easier to write: `anyhow::bail!("Worker #{id} - No task found");`

bin/liquidator/src/tx_sender.rs
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    for x in &writer.interesting_tcs {
+        if !writer.processing_tcs.contains(x) {
+            trace!("  - TCS {:?}", x);
+        }
+    }
+
+    // next tcs task to execute
+    let max_tcs_batch_size = max(1, tcs_todo / tcs_capable_workers + 1);
```
ckamm, line 186: the max() looks unneeded; and for now I think it's ok to allow the job to take all the tcs candidates, or maybe up to a fairly high static limit (20?). 

Right now, if there's 4 candidates and 4 tcs_capable_workers, it'd be split into a single candidate per worker. That makes sense generally, but avoids the jup-caching and token-modification-tracking that is the main reason the tcs batching survives this PR. 

Once those two things are resolved it'd be a large simplicity gain to drop the tcs batching.

bin/liquidator/src/tx_sender.rs
```diff
@@ -0,0 +1,237 @@
...
+            tokio::select!(
+                _ = liq_receiver.recv() => {},
+                _ = tcs_receiver.recv() => {},
+            )
+        };
+
+        // a task must be available to process
+        // find it in global shared state, and mark it as processing
+        let task = worker_pull_task(&shared_state, id, tcs_capable_workers, only_liquidation)
+            .expect("Worker woke up but has nothing to do");
```
ckamm, line 106: I think with the tcs batching pulled back in, this no longer works. You send a trigger for each tcs candidate, but the processing tasks eat up multiple at once - so for N candidates the there'll be fewer than N worker tasks.

There's also some weirdness now where the tcs jobs trigger on the tcs channel and take all the liquidation jobs and then the liquidation job wakes up (on the liq channel) and has no work to do.

It may be more robust to avoid the requirement that the number of messages on the channels and the number of work chunks match up exactly.

farnyser: It "works" because the TCS task will just be made of an empty vector (no candidate) if everything was consumed before

ckamm: Ah, I see! Would you be up for making it more explicit by adding a no-work WorkerTask?

ckamm (approved):

PR #917, opened 2024-03-14
Lou/ts fallback fetching

Lou-Kamades:


ts/client/src/client.ts
```diff
@@ -5206,9 +5222,80 @@ export class MangoClient {
...
+          .forEach((b) => {
+            oracles.push(b[0].oracle);
+            fallbacks.push(b[0].fallbackOracle);
+          });
+        return createFallbackOracleMap(this.connection, oracles, fallbacks);
+      }
+      case 'all': {
+        const oracles: PublicKey[] = [];
+        const fallbacks: PublicKey[] = [];
+        Array.from(group.banksMapByTokenIndex.values()).forEach((b) => {
```
ckamm, line 5287: imo this should only include the fallbacks for the oracles that are used by the mangoAccounts passed to buildHealthRemainingAccounts(). Fallbacks for all listed tokens might just blow the account limit.

ts/client/src/accounts/oracle.ts
```diff
@@ -186,3 +236,50 @@ export function isOracleStaleOrUnconfident(
...
+  }
+}
+// Assumes oracles.length === fallbacks.length
+export async function createFallbackOracleMap(
+  conn: Connection,
+  oracles: PublicKey[],
+  fallbacks: PublicKey[],
+): Promise<Map<PublicKey, [PublicKey, PublicKey]>> {
+  const map: Map<PublicKey, [PublicKey, PublicKey]> = new Map();
+  const accounts = await conn.getMultipleAccountsInfo(fallbacks);
```
ckamm, line 273: it's a bit sad that this is going to be re-fetched so often when it's almost never going to change

PR #919, opened 2024-03-20
Liquidator: add Sanctum swap

farnyser:


lib/client/src/swap/sanctum.rs
```diff
@@ -0,0 +1,306 @@
...
+            .mango_client
+            .http_client
+            .post(format!("{}/swap", config.sanctum_url))
+            .json(&SwapRequest {
+                amount: quote
+                    .in_amount
+                    .clone()
+                    .expect("sanctum require a in amount")
+                    .clone(),
+                quoted_amount: quote.out_amount.clone(),
```
ckamm, line 153: how do you know what to pass as amount / quoted_amount?

lib/client/src/swap/sanctum.rs
```diff
@@ -0,0 +1,306 @@
...
+            - sanctum_ixs
+                .iter()
+                .rev()
+                .position(|ix| !is_setup_ix(ix.program_id))
+                .unwrap();
+
+        let mut instructions: Vec<Instruction> = Vec::new();
+
+        for ix in &sanctum_ixs[..sanctum_action_ix_begin] {
+            instructions.push(ix.clone());
```
ckamm, line 206: this is potentially pushing SOL -> wSOL wrapping ix

lib/client/src/swap/sanctum.rs
```diff
@@ -0,0 +1,368 @@
...
+            .context("quote request to sanctum")?;
+        let quote: QuoteResponse =
+            util::http_error_handling(response).await.with_context(|| {
+                format!("error requesting sanctum route between {input_mint} and {output_mint} (using url: {})", config.sanctum_url)
+            })?;
+
+        Ok(quote)
+    }
+
+    /// Find the instructions and account lookup tables for a jupiter swap through mango
```
microwavedcola1, line 97: nit: sanctum swap and not jupiter

lib/client/src/swap/sanctum.rs
```diff
@@ -0,0 +1,368 @@
...
+            .timeout(self.timeout_duration)
+            .send()
+            .await
+            .context("swap transaction request to sanctum")?;
+
+        let swap_r: SanctumSwapResponse = util::http_error_handling(swap_response)
+            .await
+            .context("error requesting sanctum swap")?;
+
+        // let tx: SanctumSwapInstructionsResponse = SanctumSwapInstructionsResponse::deserialize(tx.as_slice())?;
```
microwavedcola1, line 184: nit: remove dead code

lib/client/src/swap/sanctum.rs
```diff
@@ -0,0 +1,368 @@
...
+        let token_program = anchor_spl::token::ID;
+        let compute_budget_program: Pubkey = solana_sdk::compute_budget::ID;
+        // these setup instructions should be placed outside of flashloan begin-end
+        let is_setup_ix = |k: Pubkey| -> bool {
+            k == ata_program || k == token_program || k == compute_budget_program
+        };
+        let sync_native_pack =
+            anchor_spl::token::spl_token::instruction::TokenInstruction::SyncNative.pack();
+
+        // Remove auto wrapping of SOL->mSOL
```
microwavedcola1, line 209: nit: I think you mean sol -> wsol or?

bin/liquidator/src/rebalance.rs
```diff
@@ -147,29 +159,34 @@ impl Rebalancer {
...
+                self.swap_quote(alt_mint, output_mint, alt_in_amount, true, jupiter_version);
             jobs.push(direct_route_job);
         }
 
-        let mut results = futures::future::join_all(jobs).await;
-        let full_route = results.remove(0)?;
-        let alternatives = results.into_iter().filter_map(|v| v.ok()).collect_vec();
+        if self.config.use_sanctum && self.is_lst(output_mint) {
+            // go through an alternative token
+            for in_token_index in &self.config.alternate_sanctum_route_tokens {
```
ckamm, line 171: When enabled, I'd like to see sanctum used also when in and out mints are LSTs, even when alternate_sanctum_route_tokens is empty.

bin/liquidator/src/rebalance.rs
```diff
@@ -147,29 +159,34 @@ impl Rebalancer {
...
+                    output_mint,
+                    alt_in_amount,
+                    true,
+                    swap::Version::Sanctum,
+                );
+                jobs.push(sanctum_alt_route_job);
+            }
+        }
+
+        let results = futures::future::join_all(jobs).await;
```
ckamm, line 185: future pr: instead of configuring individual timeouts, it could be good to give a general quote timeout that applies here

bin/liquidator/src/rebalance.rs
```diff
@@ -565,4 +638,42 @@ impl Rebalancer {
...
+        self.lst_mints.contains(&mint)
+    }
+
+    pub async fn init(&mut self, live_rpc_client: &RpcClient) {
+        if let Err(e) = self.load_lst_if_needed(live_rpc_client).await {
+            warn!("Could not load list of sanctum supported mint: {}", e);
+        }
+    }
+
+    async fn load_lst_if_needed(&mut self, live_rpc_client: &RpcClient) -> anyhow::Result<()> {
```
ckamm, line 652: nit: the `_if_needed` doesn't seem needed. Make a function that loads LST mints and returns them, then use it in init() instead of mutating `self` here?

bin/liquidator/src/rebalance.rs
```diff
@@ -565,4 +638,42 @@ impl Rebalancer {
...
+        }
+
+        let address = Pubkey::from_str("EhWxBHdmQ3yDmPzhJbKtGMM9oaZD42emt71kSieghy5")?;
+
+        let lookup_table_data = live_rpc_client.get_account(&address).await?;
+        let lookup_table = AddressLookupTable::deserialize(&lookup_table_data.data())?;
+        let accounts: Vec<Account> = live_rpc_client
+            .get_multiple_accounts(&lookup_table.addresses)
+            .await?
+            .drain(..)
```
ckamm, line 664: wouldn't iter() do the same thing?

bin/liquidator/src/rebalance.rs
```diff
@@ -565,4 +638,42 @@ impl Rebalancer {
...
+        if !self.lst_mints.is_empty() {
+            return Ok(());
+        }
+
+        let address = Pubkey::from_str("EhWxBHdmQ3yDmPzhJbKtGMM9oaZD42emt71kSieghy5")?;
+
+        let lookup_table_data = live_rpc_client.get_account(&address).await?;
+        let lookup_table = AddressLookupTable::deserialize(&lookup_table_data.data())?;
+        let accounts: Vec<Account> = live_rpc_client
+            .get_multiple_accounts(&lookup_table.addresses)
```
ckamm, line 662: does this emit a single GetMultipleAccounts call? those usually have an upper limit of usually 100 accounts to be fetched at once. Since ALTs can have 255 accounts, this needs to call a helper that splits it up into multiple rpc calls (maybe this function already does it, not sure)

bin/liquidator/src/rebalance.rs
```diff
@@ -222,47 +249,93 @@ impl Rebalancer {
...
-                full.output_mint
+    async fn determine_best_swap_tx(
+        &self,
+        mut routes: Vec<swap::Quote>,
+        input_mint: Pubkey,
+        output_mint: Pubkey,
+    ) -> anyhow::Result<(TransactionBuilder, swap::Quote)> {
+        let mut prices = HashMap::<Pubkey, I80F48>::new();
+        let mut get_or_fetch_price = |m| {
+            prices.get(&m).copied().unwrap_or_else(|| {
```
ckamm, line 276: the more idiomatic way to write this would be with `prices.entry(..)`.

this optimization could also be somewhat premature, since fetch_bank_price shouldn't be very expensive.

bin/liquidator/src/rebalance.rs
```diff
@@ -222,47 +249,93 @@ impl Rebalancer {
...
+                prices.insert(m, p);
+                p
+            })
+        };
+
+        routes.sort_by_cached_key(|r| {
+            let in_price = get_or_fetch_price(r.input_mint);
+            let out_price = get_or_fetch_price(r.output_mint);
+            let amount = out_price * I80F48::from_num(r.out_amount)
+                - in_price * I80F48::from_num(r.in_amount);
```
ckamm, line 295: mmh, this is not a sufficient criterion now that you've removed the favoring of the primary route

say someone wants to swap mSOL -> KIN. Due to the configuration we also check the alternate route mSOL -> SOL. This latter route is near guaranteed to have better `amount` than the one to KIN. But it will still often do worse if you factor in the SOL -> KIN second step that's needed afterwards.

suggestion: two step approach?
- fetch the routes between the correct in and out mints first (including sanctum if enabled and it's both LSTs) and find the best route (by `amount`) that fits in a transaction
- if this fails, fetch the alternate routes and grab the best one of those the same way

that also reduces the number of quote requests sent to the swap endpoints (being rate limited is a real concern)

lib/client/src/swap/sanctum.rs
```diff
@@ -0,0 +1,334 @@
...
+
+        let config = self.mango_client.client.config();
+
+        let in_amount = quote
+            .in_amount
+            .clone()
+            .expect("sanctum require a in amount");
+        let in_amount_u64 = in_amount.parse::<u64>()?;
+        let out_amount =
+            ((in_amount_u64 as f64) * (1.0 - (max_slippage_bps as f64) / 10_000.0)).ceil() as u64;
```
ckamm, line 159: the out amount depending on the in amount looks odd to me. If you convert 1 SOL into jitoSOL you'll only get 0.9 jitoSOL or sth

it's actually unclear what we want here. do we want slippage from quoted amount or slippage from oracle? I think jupiter currently does slippage from quoted amount, so this could be the same (even if it's not actually what rebalancers care about?)

in that case, this should apply the slippage to the quote.out_amount, I think?

lib/client/src/swap/sanctum.rs
```diff
@@ -0,0 +1,334 @@
...
+        let token_program = anchor_spl::token::ID;
+        let compute_budget_program: Pubkey = solana_sdk::compute_budget::ID;
+        // these setup instructions should be placed outside of flashloan begin-end
+        let is_setup_ix = |k: Pubkey| -> bool {
+            k == ata_program || k == token_program || k == compute_budget_program
+        };
+        let sync_native_pack =
+            anchor_spl::token::spl_token::instruction::TokenInstruction::SyncNative.pack();
+
+        // Remove auto wrapping of SOL->wSOL
```
ckamm, line 207: did you test a sanctum swap wSOL -> SOL to see if this also takes care of dropping unwrapping instructinos?

bin/liquidator/src/rebalance.rs
farnyser: Isn't the quote_token hardcoded to USDC ? Or do you mean in case of a future evolution ?

bin/liquidator/src/rebalance.rs
farnyser: `iter()` give a `&Option<T>` and I want a `Option<T>`

bin/liquidator/src/rebalance.rs
farnyser: It does only one call, I'll try to find the helper if it already exists, or add one :+1:

lib/client/src/swap/sanctum.rs
farnyser: Seems OK for both a swap to and from SOL (seen together)

bin/liquidator/src/rebalance.rs
```diff
@@ -565,4 +703,82 @@ impl Rebalancer {
...
+    }
+
+    pub async fn init(&mut self, live_rpc_client: &RpcClient) {
+        match self.load_lst(live_rpc_client).await {
+            Err(e) => warn!("Could not load list of sanctum supported mint: {}", e),
+            Ok(lst) => self.lst_mints.extend(lst),
+        }
+    }
+
+    async fn load_lst(&mut self, live_rpc_client: &RpcClient) -> anyhow::Result<HashSet<Pubkey>> {
```
ckamm, line 718: this does not even need to take a `self` argument anymore

bin/liquidator/src/rebalance.rs
ckamm: hmm, into_iter() then? :)

lib/client/src/gpa.rs
```diff
@@ -148,3 +149,44 @@ pub async fn fetch_multiple_accounts(
...
+                    Err(e) => Err(e),
+                }
+            }
+        })
+        .buffer_unordered(parallel_rpc_requests)
+        .collect::<Vec<_>>()
+        .await;
+
+    let result = raw_result
+        .drain(..)
```
ckamm, line 185: into_iter

lib/client/src/gpa.rs
```diff
@@ -148,3 +149,44 @@ pub async fn fetch_multiple_accounts(
...
+        .map(|v| v.unwrap())
```
ckamm, line 186: This unwrap() would presumably panic on rpc timeouts? instead, the above .collect() could be `.collect::<Result<Vec<_>>>` which propagates the first inner error outwards, so you could just `?` on after the collect.

lib/client/src/gpa.rs
```diff
@@ -148,3 +149,44 @@ pub async fn fetch_multiple_accounts(
...
+        .flatten()
+        .filter_map(|x| x.1.map(|o| (x.0, o)))
```
ckamm, line 188: prefer: `.filter_map(|(pubkey, account_opt)| account_opt.map(|acc| (pubkey, acc)))`  (but even that needs some thought to read, unfortunately)

Is it even acceptable to not return data for some of the accounts that were requested? Could you put in the doc comment that this is could return no data without an error?

bin/liquidator/src/rebalance.rs
```diff
@@ -146,123 +160,247 @@ impl Rebalancer {
...
+                let token = self
+                    .mango_client
+                    .context
+                    .token_by_mint(&m)
+                    .expect("token for mint not found");
+                let p = self
+                    .account_fetcher
+                    .fetch_bank_price(&token.first_bank())
+                    .expect("failed to fetch price");
+                p
```
ckamm, line 350: very minor: could drop the p binding and just let it be the last expression

bin/liquidator/src/rebalance.rs
```diff
@@ -644,36 +645,76 @@ impl Rebalancer {
...
-            .get_multiple_accounts(&lookup_table.addresses)
-            .await?
-            .drain(..)
-            .filter_map(|x| x)
-            .collect();
-
+        let accounts: Vec<Account> =
+            fetch_multiple_accounts_in_chunks(live_rpc_client, &lookup_table.addresses, 100, 1)
+                .await?
+                .drain(..)
```
ckamm, line 726: as usual, don't like the drain()

bin/liquidator/src/rebalance.rs
```diff
@@ -266,6 +290,45 @@ impl Rebalancer {
...
     }
 
+    fn can_use_sanctum_for_token(&self, mint: Pubkey) -> anyhow::Result<bool> {
+        if !self.config.use_sanctum {
+            return Ok(false);
+        }
+
+        let token = self.mango_client.context.token_by_mint(&mint)?;
+
+        let is_lst = self.is_lst(mint);
```
ckamm, line 300: since it returns true for SOL, this isn't about lst-ness, rename to sth like `is_sanctum_supported()`

bin/liquidator/src/rebalance.rs
```diff
@@ -266,6 +290,45 @@ impl Rebalancer {
...
+        &self,
+        quote_fetcher: impl Fn(&u16) -> anyhow::Result<(Pubkey, Pubkey, u64, bool)>,
+        output_mint: Pubkey,
+        jupiter_version: swap::Version,
+        quote_mint: Pubkey,
+    ) -> anyhow::Result<(TransactionBuilder, swap::Quote)> {
+        let mut alt_jobs = vec![];
+        for in_token_index in &self.config.alternate_jupiter_route_tokens {
+            let args = quote_fetcher(in_token_index)?;
+            alt_jobs.push(self.swap_quote(args.0, args.1, args.2, args.3, jupiter_version));
```
ckamm, line 321: hmm, I wonder how this could become easier to read.

maybe: `let (in, out, amount, direct) = quote_fetcher(...)?`

Or, maybe preferably, the self.swap_quote() call could be part of the quote_fetcher function?

bin/liquidator/src/rebalance.rs
```diff
@@ -266,6 +290,45 @@ impl Rebalancer {
...
+        }
+        let alt_results = futures::future::join_all(alt_jobs).await;
+        let alt_routes: Vec<_> = alt_results.into_iter().filter_map(|v| v.ok()).collect_vec();
+
+        let best_route = self
+            .determine_best_swap_tx(alt_routes, quote_mint, output_mint)
```
ckamm, line 327: I was really confused for a while because I didn't see how quote_mint/output_mint would be useful here. Looks like the function doesn't even use them - remove

lib/client/src/gpa.rs
```diff
@@ -148,3 +149,44 @@ pub async fn fetch_multiple_accounts(
...
+        .map(|keys| {
+            let account_info_config = config.clone();
+            async move {
+                let mut keys = keys.iter().map(|x| **x).collect::<Vec<Pubkey>>();
+                let req_res = rpc
+                    .get_multiple_accounts_with_config(&keys, account_info_config)
+                    .await;
+
+                match req_res {
+                    Ok(v) => Ok(keys.drain(..).zip(v.value).collect::<Vec<_>>()),
```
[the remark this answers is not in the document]
ckamm: minor: this possibly got overlooked

lib/client/src/gpa.rs
ckamm: I'd appreciate the comment

ckamm (approved):

PR #921, opened 2024-03-25
Liquidator: randomly select token/perps for rebalancing

farnyser:
randomly select token/perps for rebalancing to avoid failing at every try if one token is having an issue

### DIFF
--- a/bin/liquidator/src/rebalance.rs
+++ b/bin/liquidator/src/rebalance.rs
@@ -69,9 +69,19 @@ impl Rebalancer {
             "checking for rebalance"
         );
 
-        self.rebalance_perps().await?;
-        self.rebalance_tokens().await?;
+        let rebalance_perps_res = self.rebalance_perps().await;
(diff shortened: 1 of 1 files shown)


microwavedcola1 (approved):

bin/liquidator/src/rebalance.rs
```diff
@@ -69,9 +69,19 @@ impl Rebalancer {
...
+        if rebalance_perps_res.is_err() && rebalance_tokens_res.is_err() {
+            anyhow::bail!(
+                "Failed to rebalance perps ({}) and tokens ({})",
+                rebalance_perps_res.unwrap_err(),
+                rebalance_tokens_res.unwrap_err()
+            )
+        }
+
+        rebalance_perps_res.expect("rebalancing perps failed");
+        rebalance_tokens_res.expect("rebalancing tokens failed");
```
ckamm, line 84: this means it'll panic if only the token or only the perp rebalancing fails, I'd prefer to use `_res?` here

PR #924, opened 2024-03-27
liquidator: forcefully exit process if snapshot job die

farnyser:


microwavedcola1 (approved):

lib/client/src/snapshot_source.rs
```diff
@@ -260,4 +260,10 @@ pub fn start(config: Config, mango_oracles: Vec<Pubkey>, sender: async_channel::
             };
         }
     });
+
+    tokio::spawn(async move {
+        let res = snapshot_job.await;
+        tracing::error!("Snapshot job exited, terminating process.. ({:?})", res);
+        std::process::exit(-1);
```
ckamm, line 267: While this works technically, the previous pattern was to collect all the essential jobs in main and do the "something aborted, exit" logic there. I'd prefer that to be consistent.

Secondly, while this addresses the issue with the snapshot job dying unnoticed, the underlying issue is the .expect() on the result of the get_epoch_info rpc call, as far as I remember. Since it's only affecting startup, I'm ok with the restart-to-retry approach for now though.

ckamm (approved):

PR #926, opened 2024-03-27
liquidator: add a sequence check in rebalancing

farnyser:


lib/client/src/client.rs
```diff
@@ -610,6 +610,34 @@ impl MangoClient {
...
+                            owner: mango_account.fixed.owner,
+                        },
+                        None,
+                    )
+                },
+                data: anchor_lang::InstructionData::data(&mango_v4::instruction::SequenceCheck {
+                    expected_sequence_number: mango_account.fixed.sequence_number,
+                }),
+            }],
+            self.context.compute_estimates.cu_per_mango_instruction,
```
ckamm, line 636: this is likely a gross overestimate, I'd expect the sequence check should be <10k CU

lib/client/src/client.rs
```diff
@@ -2459,6 +2487,30 @@ impl TransactionBuilder {
         Ok(signature)
     }
 
+    pub async fn send_and_confirm_with_sequence_check(
```
ckamm, line 2490: I'm not happy with putting this function into TransactionBuilder
- currently it never needs to work with MangoClient or MangoAccountValue, it's an isolated generic utility
- adding the ix to the raw instruction list at this point means that it's cu budget isn't covered by the compute budget instruction the user may have added (typically via PreparedInstructions)

Imo it's better for users to add the sequence check to their PreparedInstructions

bin/liquidator/src/rebalance.rs
```diff
@@ -493,6 +509,23 @@ impl Rebalancer {
...
+                instructions: ixs.to_instructions(),
+                signers: vec![self.mango_client.owner.clone()],
+                ..self.mango_client.transaction_builder().await?
+            };
+
+            let seq_check_ix = self
+                .mango_client
+                .sequence_check_instruction(&self.mango_account_address, account)
+                .await?;
+            tx_builder.append(seq_check_ix);
```
ckamm, line 523: minor: In cases where we have prepared instructions (ixs) I think it's preferable to add the seq-check-ix to that, to make its cu budget estimate be included when .to-instructions() is called

ckamm (approved):

PR #937, opened 2024-04-08
keeper: add CU estimate for perp update funding, perp consume events and token update index & rates

farnyser:


bin/keeper/src/crank.rs
```diff
@@ -238,11 +245,14 @@ pub async fn loop_update_index_and_rate(
                 continue;
             }
 
-            instructions.push(ix);
+            warn!(token.name, "simulation OK: ");
```
ckamm, line 248: probably drop warn

lib/client/src/context.rs
```diff
@@ -150,6 +154,10 @@ impl Default for ComputeEstimates {
             // measured around 8k, see test_basics
             cu_for_sequence_check: 10_000,
             cu_per_associated_token_account_creation: 21_000,
+            cu_perp_update_funding: 30_000,
+            cu_perp_consume_events_base: 5_000,
+            cu_perp_consume_events_per_event: 15_000,
+            cu_token_update_index_and_rates: 75_000,
```
ckamm, line 160: I suggest 40k, 10k, 18k, 90k for even more buffer; I don't trust our monitoring to raise the issue early if some token consistently can't be cranked because of too-tight cu limits

ckamm (approved):

PR #938, opened 2024-04-09
liquidator: rebalance with openbook (limit order)

farnyser:


bin/liquidator/src/rebalance.rs
```diff
@@ -298,6 +307,7 @@ impl Rebalancer {
     }
 
     async fn rebalance_tokens(&self) -> anyhow::Result<()> {
+        self.settle_and_close_all_openbook_orders().await?;
```
ckamm, line 310: Might doing a cancel+settle of all order on each end up too frequent? I guess due to the sleep workaround in the rebalance_job it's at most once every 10s.

lib/client/src/client.rs
```diff
@@ -1162,6 +1162,17 @@ impl MangoClient {
...
+        self.send_and_confirm_owner_tx(vec![ix]).await
+    }
+
+    pub fn serum3_settle_funds_instruction(
+        &self,
+        s3: &Serum3MarketContext,
+        base: &TokenContext,
+        quote: &TokenContext,
+        open_orders: Pubkey,
+    ) -> Instruction {
```
ckamm, line 1175: this should return a PreparedInstruction. Could start with cu_per_mango_instruction (even though settlement is probably cheaper than that) (could also be a separate PR)

lib/client/src/client.rs
```diff
@@ -691,7 +691,7 @@ impl MangoClient {
                     }),
                 },
             ],
-            self.instruction_cu(health_cu),
+            self.instruction_cu(health_cu) + self.context.compute_estimates.cu_per_associated_token_account_creation,
```
ckamm, line 694: :+1:

bin/liquidator/src/rebalance.rs
```diff
@@ -423,6 +409,301 @@ impl Rebalancer {
...
+        }
+        Ok(())
+    }
+
+    async fn settle_and_close_openbook_orders(
+        &self,
+        account: &Box<MangoAccountValue>,
+        token: &TokenContext,
+        market_index: &Serum3MarketIndex,
+        market: &Serum3MarketContext,
```
ckamm, line 590: separate thought: we should add the market_index to the Serum3MarketContext, TokenContext also contains the token_index, it's convenient

bin/liquidator/src/rebalance.rs
```diff
@@ -423,6 +409,301 @@ impl Rebalancer {
...
+        let open_orders_opt = account
+            .serum3_orders(*market_index)
+            .map(|x| x.open_orders)
+            .ok();
+
+        if open_orders_opt.is_none() {
+            return Ok(());
+        }
+
+        let open_orders = open_orders_opt.unwrap();
```
ckamm, line 602: nit: You could write this as
```
let Ok(open_orders) = account.s3o().map(...)
else {
    return Ok(());
};
```

bin/liquidator/src/rebalance.rs
```diff
@@ -423,6 +409,301 @@ impl Rebalancer {
...
+            return Ok(());
+        }
+
+        let settle_ixs = PreparedInstructions::from_single(
+            self.mango_client
+                .serum3_settle_funds_instruction(market, token, quote, open_orders),
+            self.mango_client
+                .context
+                .compute_estimates
+                .cu_per_serum3_order_cancel, // TODO FAS
```
ckamm, line 617: as stub I'd use the generic "mango ix cu" value

bin/liquidator/src/rebalance.rs
```diff
@@ -423,6 +409,301 @@ impl Rebalancer {
...
+        );
+
+        let close_ixs = self
+            .mango_client
+            .serum3_close_open_orders_instruction(*market_index);
+
+        let mut ixs = PreparedInstructions::new();
+        ixs.append(settle_ixs);
+        ixs.append(close_ixs);
```
ckamm, line 626: the ordering is odd, you should close open orders (making reserved funds free) and then settle (transfering the free funds into the account)

bin/liquidator/src/rebalance.rs
```diff
@@ -423,6 +409,301 @@ impl Rebalancer {
...
+            .context
+            .serum3_markets
+            .iter()
+            .find(|(_, context)| {
+                context.base_token_index == token.token_index
+                    && context.quote_token_index == QUOTE_TOKEN_INDEX
+            })
+            .ok_or(anyhow::format_err!(
+                "could not find market for token {}",
+                token.name
```
ckamm, line 424: sounds like this would be spammed a lot when there's a token that doesn't have a USDC market

bin/liquidator/src/rebalance.rs
```diff
@@ -324,7 +334,11 @@ impl Rebalancer {
             // to sell them. Instead they will be withdrawn at the end.
             // Purchases will aim to purchase slightly more than is needed, such that we can
             // again withdraw the dust at the end.
-            let dust_threshold = I80F48::from(2) / token_price;
+            let dust_threshold = if self.config.use_limit_order {
+                I80F48::from(self.dust_threshold_for_limit_order(token).await?)
```
ckamm, line 338: it's possible for this (and the unwind call later) to fail consistently if there is no available market for the token. I think a failure here should not prevent rebalancing of other tokens

bin/liquidator/src/rebalance.rs
```diff
@@ -423,6 +409,301 @@ impl Rebalancer {
...
+        };
+
+        let distance_from_oracle_price_bp =
+            I80F48::from_num(self.config.limit_order_distance_from_oracle_price_bps)
+                * match side {
+                    Serum3Side::Bid => 1,
+                    Serum3Side::Ask => -1,
+                };
+        let price_adjustment_factor =
+            (I80F48::from_num(10_000) + distance_from_oracle_price_bp) / I80F48::from_num(10_000);
```
ckamm, line 469: personally I prefer to store these things as a fraction in f64, instead of bps, e.g. `1.0 + relative_distance_from_oracle_price` - but this is ok too

bin/liquidator/src/rebalance.rs
```diff
@@ -423,6 +409,301 @@ impl Rebalancer {
...
+            side = match side {
+                Serum3Side::Bid => "Buy",
+                Serum3Side::Ask => "Sell",
+            },
+            token = token.name,
+            oracle_price = token_price.to_num::<f64>(),
+            price_adjustment_factor = price_adjustment_factor.to_num::<f64>(),
+            coin_lot_size = market.coin_lot_size,
+            pc_lot_size = market.pc_lot_size,
+            limit_price = limit_price,
```
ckamm, line 488: fyi, for convenience you could say just `limit_price,` here and it'd be the same thing

bin/liquidator/src/rebalance.rs
```diff
@@ -423,6 +409,301 @@ impl Rebalancer {
...
+            .serum3_place_order_instruction(
+                &account,
+                *market_index,
+                side,
+                limit_price,
+                max_base_lots,
+                ((limit_price * max_base_lots * market.pc_lot_size) as f64 * 1.01) as u64,
+                Serum3SelfTradeBehavior::CancelProvide,
+                if self.config.limit_order_distance_from_oracle_price_bps == 0 {
+                    Serum3OrderType::PostOnly
```
ckamm, line 527: can you explain?

bin/liquidator/src/rebalance.rs
```diff
@@ -423,6 +409,301 @@ impl Rebalancer {
...
+                } else {
+                    Serum3OrderType::Limit
+                },
+                SystemTime::now().duration_since(UNIX_EPOCH)?.as_millis() as u64,
```
ckamm, line 531: it'd be kind of cool if the liquidator wouldn't cancel and replace user orders that the owner of the account added to manually rebalance. But that ends up a bit difficult because we can no longer just cancel all. It'd probably work by using special client_order_ids. Maybe it'd even work to use just a fixed id? Not sure if cancel_by_client_order_id cancels all of that id or just the first.

Don't think you need to worry about it for this PR, just a thought.

bin/liquidator/src/rebalance.rs
farnyser: It will not do anything if there is no open orders account or if there is an active order for the token

bin/liquidator/src/rebalance.rs
farnyser: I wanted to make the liquidator only be on the maker side when configured to execute at oracle price, but it would be a hidden feature and could lead to unexpected behavior (no exec even if price would match). I'll revert that.

bin/liquidator/src/rebalance.rs
farnyser: Yeah I also though it could be good to not cancel orders if posting at the same price, but then I though this should be enough (and better to avoid unnecessary complexity)

If configured to place order at oracle price with a 3% "slippage", we should really be executed quickly enough that it does not matter.  

What do you think ? Fine to leave it like this ?

bin/liquidator/src/rebalance.rs
```diff
@@ -476,10 +476,19 @@ impl Rebalancer {
...
             } else {
-                I80F48::from(2) / token_price
+                Ok(I80F48::from(2) / token_price)
+            };
+
+            let Ok(dust_threshold) = dust_threshold_res
+            else {
+                let e = dust_threshold_res.unwrap_err();
+                error!("Cannot rebalance token {}, probably missing USDC market ? - error: {}", token.name, e);
+                return Ok(());
```
ckamm, line 491: hmm, did you mean to `continue`?

bin/liquidator/src/rebalance.rs
ckamm: Fine to leave

bin/liquidator/src/rebalance.rs
farnyser: yes sorry, fix pushed

ckamm (approved):

PR #943, opened 2024-04-16
Implement fallback oracles for perps

Lou-Kamades:


programs/mango-v4/src/instructions/perp_place_order.rs
```diff
@@ -31,9 +31,18 @@ pub fn perp_place_order(
...
+                .map(|k| AccountInfoRef::borrow(k).unwrap())
+        } else {
+            None
+        };
+
         let oracle_ref = &AccountInfoRef::borrow(ctx.accounts.oracle.as_ref())?;
         let oracle_state = perp_market.oracle_state(
-            &OracleAccountInfos::from_reader(oracle_ref),
+            &OracleAccountInfos::from_reader_with_fallback(oracle_ref, fallback_opt.as_ref()),
             None, // staleness checked in health
```
ckamm, line 46: note that continuing to pass None here will not do the staleness check and thus potentially not use the fallback when it should

programs/mango-v4/src/state/oracle.rs
```diff
@@ -295,6 +295,15 @@ impl<'a, T: KeyedAccountReader> OracleAccountInfos<'a, T> {
             sol_opt: None,
         }
     }
+
+    pub fn from_reader_with_fallback(acc_reader: &'a T, fallback_opt: Option<&'a T>) -> Self {
+        OracleAccountInfos {
+            oracle: acc_reader,
+            fallback_opt,
+            usdc_opt: None,
+            sol_opt: None,
```
ckamm, line 304: so when this function is used, raydium/orca fallbacks are not available?

ckamm (commented):
While this approach could work with some fixes (to enable raydium/orca fallbacks), my opinion is that we should stop reading the explicitly-passed oracles where possible.

Instead, we should read the oracle price from the AccountRetriever, to have a unified way of accessing oracle prices for a token/perp.

Unfortunately that will often require significant instruction adjustments (because creating the AccountRetriever is reasonably expensive and also holds a bunch of borrows)

MANAALHARBI (approved):

PR #964, opened 2024-05-22
Collateral fees only should not get worse when you add more collateral

brittcyr:


programs/mango-v4/src/instructions/token_charge_collateral_fees.rs
```diff
@@ -73,15 +73,35 @@ pub fn token_charge_collateral_fees(ctx: Context<TokenChargeCollateralFees>) ->
...
+    // collateral without any new liabilities. The program should not penalize a
+    // user for depositing more collateral.
+    let mut collateral_fee_per_day_and_bank_ai_index = Vec::with_capacity(token_position_count);
+    for index in 0..token_position_count {
+        let bank_ai = &ctx.remaining_accounts[index];
+
+        // Could directly get the bytes from the bank since they are in a fixed
+        // position and save CU, but for code readability, deserialize the whole
+        // account.
+        let bank = bank_ai.load::<Bank>()?;
```
ckamm, line 93: .load::<Bank>() is a zero-copy load, so very low CU

programs/mango-v4/src/instructions/token_charge_collateral_fees.rs
```diff
@@ -93,7 +113,24 @@ pub fn token_charge_collateral_fees(ctx: Context<TokenChargeCollateralFees>) ->
             continue;
         }
 
-        let fee = token_balance * scaling * I80F48::from_num(bank.collateral_fee_per_day);
+        let token_balance = token_position.native(&bank);
```
ckamm, line 116: read the balance from the token_balances array here - position.native() doesn't include reserved serum/obv2 amounts, perp amounts etc

PR #969, opened 2024-06-04
Cancel openbook v2 order by client order id

brittcyr:
Allow users to cancel orders to openbook v2 using the client order id so they dont have to keep track of exchange order ids


programs/mango-v4/src/instructions/openbook_v2_cancel_order_by_client_order_id.rs
```diff
@@ -0,0 +1,108 @@
...
+// cpi for cancelling by client order id. Uses the same accounts.
+pub fn openbook_v2_cancel_order_by_client_order_id(
+    ctx: Context<OpenbookV2CancelOrder>,
+    side: OpenbookV2Side,
+    client_order_id: u64,
+) -> Result<()> {
+    // Check instruction gate. Guarded by the same as cancelling by exchange id.
+    let group = ctx.accounts.group.load()?;
+    require!(
+        group.is_ix_enabled(IxGate::OpenbookV2CancelOrder),
```
farnyser, line 25: For perp we have a dedicated ix_gate for cancel by client id, maybe add one here ?

programs/mango-v4/src/instructions/openbook_v2_cancel_order_by_client_order_id.rs
farnyser, line 1: There is about 100 lines that are the same as the `openbook_v2_cancel_order`  instruction, can we maybe refactor it to avoid that duplication ?

programs/mango-v4/src/instructions/openbook_v2_cancel_order_by_client_order_id.rs
brittcyr: could if you want. dont do it in serum3 though.
would require duplicating the account_ix file because it would be a new constraint

farnyser: ok can probably keep like this, thx

farnyser (approved):

PR #972, opened 2024-06-22
feat: switchboard on-demand init

mgild:
Switchboard On-Demand is a pull-model iteration of switchboard that is now zero cost to run oracles. Instead, using the frontend sdk, the feed user can request a price update from the switchboard network and prices will be updated as the first instruction of the user transaction.

It is now zero cost for Mango to maintain switchboard feeds through on-demand and latency concerns are eliminated.

(This PR is currently a stub)

More information at https://docs.switchboard.xyz/docs


farnyser:
Merged with #974

PR #984, opened 2024-07-22
Fix price impact for USDC pairs

riordanp:
Calculating price impact fails for pairs including USDC, which prevents TCS orders from being created in the UI. 
This change uses the available information in the API response, given all tokens are quoted in USDC there, to give correct price impacts for USDC.


riordanp:
Based on wrong branch, fixed in #990 instead

PR #986, opened 2024-08-05
Liquidator: fix an arithmetic error in client side health computation

farnyser:


riordanp (approved):
lgtm, at least does not make things worse

PR #992, opened 2024-08-12
Serge/bank token tier

farnyser:


programs/mango-v4/src/instructions/token_register_trustless.rs
```diff
@@ -111,6 +110,7 @@ pub fn token_register_trustless(
         collected_liquidation_fees: I80F48::ZERO,
         collected_collateral_fees: I80F48::ZERO,
         collateral_fee_per_day: 0.0, // TODO
+        tier: fill_from_str("C")?,
```
mschneider, line 113: please verify in test-case

ts/client/src/accounts/group.ts
```diff
@@ -359,7 +359,7 @@ export class Group {
         Array.from(this.serum3MarketsMapByExternal.values()).map(
           (serum3Market) =>
             Market.load(
-              client.program.provider.connection,
+              client.program.provider.connection as any,
```
mschneider, line 362: this is probably masking an underlying version incompatibility issue, would it be possible to just update or override the dependency?

PR #993, opened 2024-08-13
alloc fixes

mschneider:


microwavedcola1:
picking up the more important parts here https://github.com/blockworks-foundation/mango-v4/pull/1004

